PR #9, opened 2015-01-04
[JENKINS-26276] Implemented JUnitStep ("archiveTestResults")

akbertram:
Here is a first attempt at a workflow step that returns test summary results to the caller.
Can be used as follows:

```
def tests = archiveTestResults()
if(tests.failCount > 0) {
  // take action!
}
```

Feedback welcome!

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitStep.java
+++ b/src/main/java/hudson/tasks/junit/JUnitStep.java
@@ -1,0 +1,185 @@
+package hudson.tasks.junit;
+
+import hudson.*;
+import hudson.model.Result;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.remoting.VirtualChannel;
+import org.apache.tools.ant.DirectoryScanner;
+import org.apache.tools.ant.types.FileSet;
+import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousStepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.jenkinsci.remoting.RoleChecker;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
+
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.logging.Logger;
+
+/**
+ * Workflow step that archives JUnit Test results and returns a summary of counts
+ * to the caller
+ */
+public class JUnitStep extends AbstractStepImpl {
+
+    private static final Logger LOGGER = Logger.getLogger(JUnitStep.class.getName());
+
+    @Nullable
+    private String location;
+
+    private boolean keepLongStdio;
+
+    @DataBoundConstructor
+    public JUnitStep(@Nullable String location) {
+        this.location = location;
+    }
+
+    @Nullable
+    public String getLocation() {
+        return location;
+    }
+
+    public boolean isKeepLongStdio() {
+        return keepLongStdio;
+    }
+
+    @DataBoundSetter
+    public void setLocation(@Nullable String location) {
+        this.location = location;
+    }
+
+    public void setKeepLongStdio(boolean keepLongStdio) {
+        this.keepLongStdio = keepLongStdio;
+    }
+
+    public static final class ExecutionImpl extends AbstractSynchronousStepExecution<TestResultSummary> {
+
+        @Inject
+        private transient JUnitStep step;
+
+
+        @StepContextParameter
+        private transient TaskListener listener;
+
+        @StepContextParameter
+        private transient FilePath workspace;
+
+        @StepContextParameter
+        private transient EnvVars env;
+
+        @StepContextParameter
+        private transient Run build;
+
+
+        @Override
+        protected TestResultSummary run() throws Exception {
+
+            final long buildTime = build.getTimestamp().getTimeInMillis();
+            final long timeOnMaster = System.currentTimeMillis();
+
+            String location = step.location;
+            if (location == null) {
+                location = "**/target/surefire-reports/TEST-*.xml";
+            }
+
+            LOGGER.info("Looking for JUnit tests results in " + location + " in workspace " + workspace);
+
+            TestResult result = workspace.act(new ParseResultCallable(location, buildTime, timeOnMaster, step.keepLongStdio));
+            result.tally();
+
+            // Check for empty tests
+            if (result.isEmpty()) {
+                // most likely a configuration error in the job - e.g. false pattern to match the JUnit result files
+                throw new AbortException(Messages.JUnitResultArchiver_ResultIsEmpty());
+            }
+
+            // Add the results to the build
+            synchronized (build) {
+                TestResultAction action = build.getAction(TestResultAction.class);
+                if (action == null) {
+                    action = new TestResultAction(build, result, listener);
+                    build.addAction(action);
+
+                } else {
+                    // add to existing results
+                    result.freeze(action);
+                    action.mergeResult(result, listener);
+                    build.save();
+                }
+
+                if (action.getResult().getFailCount() > 0) {
+                    build.setResult(Result.UNSTABLE);
+                }
+            }
+
+            return new TestResultSummary(result);
+        }
+    }
+
+    @Extension
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+
+        public DescriptorImpl() {
+            super(ExecutionImpl.class);
+        }
+
+        @Override
+        public String getFunctionName() {
+            return "archiveTestResults";
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Archive JUnit Test Results";
+        }
+    }
+
+
+    private static final class ParseResultCallable implements FilePath.FileCallable<TestResult> {
+        private final long buildTime;
+        private final String testResults;
+        private final long nowMaster;
+        private final boolean keepLongStdio;
+
+        private ParseResultCallable(String testResults, long buildTime, long nowMaster, boolean keepLongStdio) {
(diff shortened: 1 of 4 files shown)


jglick:
The problem with doing this in `junit` is it means we cannot then bundle an update to this plugin in `jenkins.war` unless `workflow-step-api` is also bundled, which seems a big step. On the other hand the `junit` plugin is the most logical place for it.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    public boolean isKeepLongStdio() {
+        return keepLongStdio;
+    }
+
+    @DataBoundSetter
+    public void setLocation(@Nullable String location) {
+        this.location = location;
+    }
```
jglick: Delete this. If `location` is defined in a `@DataBoundConstructor`, it should be `final` and there should be no `@DataBoundSetter`.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public void setKeepLongStdio(boolean keepLongStdio) {
+        this.keepLongStdio = keepLongStdio;
+    }
```
jglick: This, on the other hand, _does_ need to be a `@DataBoundSetter`.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+public class JUnitStep extends AbstractStepImpl {
+
+    private static final Logger LOGGER = Logger.getLogger(JUnitStep.class.getName());
+
+    @Nullable
+    private String location;
+
+    private boolean keepLongStdio;
+
+    @DataBoundConstructor
```
jglick: You need a matching `config.jelly`.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+        @Override
+        protected TestResultSummary run() throws Exception {
+
+            final long buildTime = build.getTimestamp().getTimeInMillis();
+            final long timeOnMaster = System.currentTimeMillis();
+
+            String location = step.location;
+            if (location == null) {
+                location = "**/target/surefire-reports/TEST-*.xml";
```
jglick: Better just make this field mandatory, and if you like specify the `default` attribute on `f:textbox` in `config.jelly` (and/or suggest this value in `help-location.html`).

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+            LOGGER.info("Found files: " + Arrays.toString(files));
+
+            return result;
+        }
+
+        @Override
+        public void checkRoles(RoleChecker roleChecker) throws SecurityException {
+
+        }
```
jglick: Extend `MasterToSlaveFileCallable` instead.

src/main/java/hudson/tasks/junit/TestResultSummary.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public TestResultSummary(TestResult result) {
+        result.tally();
+        failCount = result.getFailCount();
+        passCount = result.getPassCount();
+        skipCount = result.getSkipCount();
+        totalCount = result.getTotalCount();
+    }
+
+    public int getFailCount() {
```
jglick: All these getters would need to be `@Whitelisted` to be accessible by default from a sandboxed script.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+                } else {
+                    // add to existing results
+                    result.freeze(action);
+                    action.mergeResult(result, listener);
+                    build.save();
+                }
+
+                if (action.getResult().getFailCount() > 0) {
+                    build.setResult(Result.UNSTABLE);
```
jglick: It is not obvious to me that it is desirable to unconditionally mark the build unstable if there are test failures. Depends on what your original use case for JENKINS-26276 really is.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+            LOGGER.info("Looking for JUnit tests results in " + location + " in workspace " + workspace);
+
+            TestResult result = workspace.act(new ParseResultCallable(location, buildTime, timeOnMaster, step.keepLongStdio));
+            result.tally();
+
+            // Check for empty tests
+            if (result.isEmpty()) {
+                // most likely a configuration error in the job - e.g. false pattern to match the JUnit result files
+                throw new AbortException(Messages.JUnitResultArchiver_ResultIsEmpty());
```
jglick: Same here. Should probably be left to script to decide what to do if `totalCount == 0`.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+            FileSet fs = Util.createFileSet(ws, testResults);
+            DirectoryScanner ds = fs.getDirectoryScanner();
+
+            TestResult result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio);
+
+            String[] files = ds.getIncludedFiles();
+            if (files.length == 0) {
+                // no test result. Most likely a configuration
+                // error or fatal problem
+                throw new AbortException(hudson.tasks.junit.Messages.JUnitResultArchiver_NoTestReportFound());
```
jglick: Another case where for Workflow it may not make sense for this to be fatal.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        @Override
+        public TestResult invoke(File ws, VirtualChannel channel) throws IOException {
+            final long nowSlave = System.currentTimeMillis();
+
+            FileSet fs = Util.createFileSet(ws, testResults);
+            DirectoryScanner ds = fs.getDirectoryScanner();
+
+            TestResult result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio);
```
jglick: Not quite sure what the time calculations here are doing (a comment would be nice), but remember that slaves may have clocks that are skewed away from those of the master.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+
+            String[] files = ds.getIncludedFiles();
+            if (files.length == 0) {
+                // no test result. Most likely a configuration
+                // error or fatal problem
+                throw new AbortException(hudson.tasks.junit.Messages.JUnitResultArchiver_NoTestReportFound());
+            }
+
+            LOGGER.info("Found files: " + Arrays.toString(files));
```
jglick: Too verbose.

src/main/java/hudson/tasks/junit/JUnitStep.java
```diff
@@ -0,0 +1,185 @@
...
+                if (action == null) {
+                    action = new TestResultAction(build, result, listener);
+                    build.addAction(action);
+
+                } else {
+                    // add to existing results
+                    result.freeze(action);
+                    action.mergeResult(result, listener);
+                    build.save();
+                }
```
jglick: I have struggled to get comparable logic in `JUnitResultArchiver` right, so this definitely needs a test (covering both branches).

src/main/java/hudson/tasks/junit/TestResultSummary.java
```diff
@@ -0,0 +1,41 @@
...
+public class TestResultSummary implements Serializable {
+
+
+    private int failCount;
+    private int passCount;
+    private int skipCount;
+    private int totalCount;
+
+    public TestResultSummary(TestResult result) {
+        result.tally();
```
jglick: `run` already called `tally`, so do you need to call it again?

akbertram:
Thanks for the feedback! Will try to submit fixes by the end of next week.

src/main/java/hudson/tasks/junit/JUnitStep.java
akbertram: I don't use the build status at all so happy to remove, but is there a way for a workflow to set the build to "unstable" if desired?

jglick: Currently only by using the standard `JUnitResultArchiver`. There is an `error` step that throws up an `AbortException` which will mark it as `FAILED` as well as terminating the flow, but perhaps there should be a special step just for setting build status without interrupting control flow or adding test results. Not sure. Another option would be a way to get the `[Workflow]Run` instance directly (either as a new step, or a predefined variable), at which point the proposed fix of [JENKINS-25851](https://issues.jenkins-ci.org/browse/JENKINS-25851) would also let that object be used in the natural way.

In the meantime, for purposes of this step, I would suggest at least making the setting of stability be optional.

omehegan:
Any plans to continue work on this?

miguelaferreira:
+1 on this feature

oleg-nenashev:
@jglick Are you fine with the current impl? If yes, I'll try to integrate it

olivergondza:
> The problem with doing this in  junit  is it means we cannot then bundle an update to this plugin in  jenkins.war  unless  workflow-step-api  is also bundled, which seems a big step.

How about make it an optional dependency? I would vote for that even if junit were not bundled...

olivergondza:
Also, the feature is untested.

abayer:
+1 for resurrecting this - it'd be a handy feature.

oleg-nenashev:
@abayer Are you working on it in #47

ziru:
Any update on this PR?  No go?

oleg-nenashev:
@ziru IIRC @abayer tried to implement something in #47, then @jglick -1'ed the approach. I suppose there is not much progress now

abayer:
Closing - I added a `junit` step (replacing the `junit` `@Symbol`) in https://github.com/jenkinsci/junit-plugin/pull/76

PR #10, opened 2015-02-03
Avoid NPE if someone calls a deprecated constructor

jglick:
Would probably fix the test failure in https://github.com/jenkinsci/flaky-test-handler-plugin/pull/1 though that is better done by fixing the test to not call this constructor to begin with. @reviewbybees

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResultAction.java
+++ b/src/main/java/hudson/tasks/junit/TestResultAction.java
@@ -81,9 +81,9 @@ public class TestResultAction extends AbstractTestResultAction<TestResultAction>
         setResult(result, listener);
     }
 
-    /** @since 1.545 */
+    @Deprecated
     public TestResultAction(TestResult result, BuildListener listener) {
-        this(null, result, listener);
+        this((Run) null, result, listener);
     }
 
     @Override public Collection<? extends Action> getProjectActions() {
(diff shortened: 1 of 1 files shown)


tfennelly:
:+1:

stephenc:
:+1:

oleg-nenashev:
LGTM

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -101,12 +101,14 @@ public synchronized void setResult(TestResult result, TaskListener listener) {
         failCount = result.getFailCount();
         skipCount = result.getSkipCount();
 
+        if (run != null) {
         // persist the data
```
jtnord: is this github or is the contained body not indented...
Yuck

jglick: Intentionally not indented, to minimize the diff.

jtnord:
For future reference:
  diff fluff is no excuse , you can add the query parameter w=1 to ignore white-space changes from showing in the diff in github.

jglick:
Yes I know about `?w=1` (which is only a workaround: you cannot for example comment on lines in that mode); but the main purpose was not to prettify the diff view, but to avoid changing source lines, which complicates `git-blame`, `git-merge`, `git-cherry-pick`, ad nauseam.

PR #12, opened 2015-02-16
Trim stdio even on failing tests, just to a 100× larger cutoff

jglick:
When `!keepLongStdio`, the stdout/stderr of passing tests is truncated to avoid memory pressure, under the assumption that you probably do not care about the details much. This is good, but it is useful to have some kind of limit even for failing tests. Otherwise a test that produces copious log output, then fails, can overwhelm the remoting channel and the Jenkins master heap. This patch sets the limit to be around 100K per incident (in JUnit, that means per test suite).

@reviewbybees (ref. ZD-24570)

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -143,39 +143,40 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         stderr = possiblyTrimStdio(_this, keepLongStdio, testCase.elementText("system-err"));
     }
 
-    private static final int HALF_MAX_SIZE = 500;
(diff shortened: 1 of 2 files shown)


tfennelly:
:+1: 

Perhaps those `possiblyTrimStdio` methods could be dried out a bit, both delegating the reading of the head and tail to a version that takes a `TextReader` version of `TextFile`. Just a thought.

recampbell:
:+1:

jglick:
> Perhaps those `possiblyTrimStdio` methods could be dried out a bit

Perhaps. I am reluctant to really change the implementation of these, since my original implementation apparently did not work as it was supposed to at all and @kohsuke had to fix it, but the problem (excess disk I/O) is difficult or impossible to test.

PR #15, opened 2015-04-15
[JENKINS-18696] Fix some tests are not reported

harcher81:
* Some tests file were not reported when test-suite name and timestamps are the same. This can happen for example if the user run the same test on 2 platforms.
* Converted tab to 4 spaces
* Added logging

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -221,7 +221,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     }
     
     public ClassResult getParent() {
-    	return classResult;
+        return classResult;
     }
 
     private static String getError(Element testCase) {
@@ -338,14 +338,14 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     
     @Override
     public String getFullName() {
-    	return className+'.'+getName();
+        return className+'.'+getName();
     }
     
     /**
      * @since 1.515
      */
     public String getFullDisplayName() {
-    	return TestNameTransformer.getTransformedName(getFullName());
+        return TestNameTransformer.getTransformedName(getFullName());
     }
 
     @Override
@@ -386,7 +386,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     }
     
     public Run<?,?> getFailedSinceRun() {
-    	return getRun().getParent().getBuildByNumber(getFailedSince());
+        return getRun().getParent().getBuildByNumber(getFailedSince());
     }
 
     /**
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -221,7 +221,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     }
     
     public ClassResult getParent() {
-        return classResult;
+    	return classResult;
     }
 
     private static String getError(Element testCase) {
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -221,7 +221,7 @@ public CaseResult(SuiteResult parent, String testName, String errorStackTrace) {
     }
     
     public ClassResult getParent() {
-    	return classResult;
+        return classResult;
```
oleg-nenashev: Please avoid formatting changes

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -95,12 +95,22 @@ public TestResult parseResult(String testResultLocations,
...
         private final long nowMaster;
         private final boolean keepLongStdio;
+        private final TaskListener listener;
 
         private ParseResultCallable(String testResults, long buildTime, long nowMaster, boolean keepLongStdio) {
             this.buildTime = buildTime;
             this.testResults = testResults;
             this.nowMaster = nowMaster;
             this.keepLongStdio = keepLongStdio;
+            this.listener = null;
```
oleg-nenashev: This constructor should call the generic one using <code>this()</code>.

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -116,9 +126,9 @@ public TestResult invoke(File ws, VirtualChannel channel) throws IOException {
                 throw new AbortException(Messages.JUnitResultArchiver_NoTestReportFound());
             }
 
-            TestResult result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio);
+            TestResult result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio, listener);
```
oleg-nenashev: Invoke is being executed on the slave's side. Most probably TaskListener won't behave as you expect

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -287,6 +287,20 @@ public SuiteResult getPreviousResult() {
         return null;
     }
 
+     /**
+     * Check if all the tests exists for the suiteResult
+     * @param suiteResult
+     * @return true if all the tests were found, false otherwise
+     */
+    public boolean includesTestsFrom(SuiteResult suiteResult) {
```
oleg-nenashev: <code>includesAllTests()</code> would be better

oleg-nenashev: Laso it makes sense to add <code>@since TODO</code> to document method in API when it gets merged

jglick:
I think this is right, but the correctness is not entirely clear, and any regressions would affect a lot of people. A test demonstrating the fix would be in order here. Should not be hard to write: you prepare a control test case where you create a suite result file with a given timestamp, and a set of matching cases with the same timestamp, and assert that you get only one copy of the results (to prove the correctness of the original fix: 0cddc1302469a0299bad838cccfedac9d53e2886); then a main test case (which you verify fails before your patch is applied) where the cases do not match the contents of the suite, and assert that nothing is dropped.

src/main/java/hudson/tasks/junit/SuiteResult.java
jglick: Or just make the method package-private. I see no particular reason to expose this into the API.

oleg-nenashev:
@harcher81 
I stink the implementation is not completely correct in general
* You check if **ALL** tests cases are present in the target test suite
* If no, you add *ALL*** tests cases to the suite => there can be dups
* <code>TestSuite.add()</code> is not aware about possible duplications

It seems to be a real risk if somebody appends test cases and resubmits the same TestResult. I would propose to implement a duplicates aware <code>TestSuite.add()</code> logic and then get rid of timestamps check at all.

jglick:
> I would propose to implement a duplicates aware `TestSuite.add()` logic and then get rid of timestamps check

Agreed, this would seem more straightforward.

harcher81:
I see, I close this pull request. I'll open a new pull request with your comments.

jglick:
There is no need to close a pull request. You can simply push new commits to the same branch and the PR gets updated.

PR #16, opened 2015-04-16
[JENKINS-27974]: Safe URL for empty className

ydubreuil:
Opening a test result with a empty className tag in junitReport.xml file results in a 404 page because the URL contains a empty path for the class.

This patch prevents this.

### DIFF
--- a/src/main/java/hudson/tasks/test/TestObject.java
+++ b/src/main/java/hudson/tasks/test/TestObject.java
@@ -373,8 +373,12 @@ public abstract class TestObject extends hudson.tasks.junit.TestObject {
      * Replaces URL-unsafe characters.
      */
     public static String safe(String s) {
-        // this still seems to be a bit faster than a single replace with regexp
-        return s.replace('/', '_').replace('\\', '_').replace(':', '_').replace('?', '_').replace('#', '_').replace('%', '_');
+        if (s.isEmpty()) {
+            return "(empty)";
+        } else {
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
The change seems to be correct, but Java classname to URL conversions stills error-prone. Java class name can contain almost any symbol (see [Java identifiers spec][1]). I would propose to use existing UNICODE=>URL conversion methods instead of string replacements.

[1]http://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8

ydubreuil:
@oleg-nenashev I did not want to change the current escaping code, so I let it like it was.

@reviewbybees

oleg-nenashev:
:+1: for this change. className escaping can be just submitted as another issue for the future use

oleg-nenashev:
BTW, you could use StringUtils.isEmpty() to check the null case as well. Also consider documenting the behaviour in Javadoc

ydubreuil:
I've written javadoc and used StringUtils.isEmpty

jglick:
:+1: 

BTW for the convenience of reviewers please hyperlink to the issue in the PR description, like so:

[JENKINS-27974](https://issues.jenkins-ci.org/browse/JENKINS-27974)

ydubreuil:
Right @jglick, will do that next time

oleg-nenashev:
I reconfirm :+1:
// Javadoc Nazi may disagree with the missing of <code>@return</code> tag, but it's just for the future

PR #17, opened 2015-04-22
[JENKINS-27931] Allow to configure the number of bytes to keep from stdout/stderr

Starch:
This patch allows to configure the maximum desired size to keep from stdout/stderr on tests.

By default the value of 1Kb (for succeeding test) and 100Kb (for failing test) are kept.

Setting a value of -1 will keep all output, 0 will keep nothing, X will truncate to X bytes, provided that keepLongStdio is false. On fact my personal preference goes to keep none for succeeding tests, and keep all for failing ones.


tfiskgul:
This looks promising, I was just looking for something like this. @Starch could you rebase this again to resolve the merge conflicts? It would be great if a maintainer could review it also.

Starch:
Just rebased (sorry for the delay).

timja:
if you / or someone else rebases this against current master I can take a look.

PR #18, opened 2015-04-29
[JENKINS-26077] Duplicated TestResultProjectAction

jglick:
[JENKINS-26077](https://issues.jenkins-ci.org/browse/JENKINS-26077)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResultAction.java
+++ b/src/main/java/hudson/tasks/junit/TestResultAction.java
@@ -24,10 +24,12 @@
 package hudson.tasks.junit;
 
 import com.thoughtworks.xstream.XStream;
+import hudson.Util;
 import hudson.XmlFile;
 import hudson.model.AbstractBuild;
 import hudson.model.Action;
 import hudson.model.BuildListener;
+import hudson.model.Job;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.test.AbstractTestResultAction;
@@ -86,8 +88,14 @@ public class TestResultAction extends AbstractTestResultAction<TestResultAction>
         this((Run) null, result, listener);
     }
 
+    @SuppressWarnings("deprecation")
     @Override public Collection<? extends Action> getProjectActions() {
-		return Collections.<Action>singleton(new TestResultProjectAction(run.getParent()));
+        Job<?,?> job = run.getParent();
+        if (/* getAction(Class) produces a StackOverflowError */!Util.filter(job.getActions(), TestResultProjectAction.class).isEmpty()) {
+            // JENKINS-26077: someone like XUnitPublisher already added one
+            return Collections.emptySet();
+        }
+		return Collections.singleton(new TestResultProjectAction(job));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -86,8 +88,14 @@ public TestResultAction(TestResult result, BuildListener listener) {
         this((Run) null, result, listener);
     }
 
+    @SuppressWarnings("deprecation")
```
oleg-nenashev: Does it really make sense to filter away deprecations? It would be better to leave them as warnings

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -86,8 +88,14 @@ public TestResultAction(TestResult result, BuildListener listener) {
...
     @Override public Collection<? extends Action> getProjectActions() {
-		return Collections.<Action>singleton(new TestResultProjectAction(run.getParent()));
+        Job<?,?> job = run.getParent();
+        if (/* getAction(Class) produces a StackOverflowError */!Util.filter(job.getActions(), TestResultProjectAction.class).isEmpty()) {
+            // JENKINS-26077: someone like XUnitPublisher already added one
+            return Collections.emptySet();
+        }
+		return Collections.singleton(new TestResultProjectAction(job));
```
oleg-nenashev: It makes sense to finally get rid of tabs, because the string has been modified in any case

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -86,8 +88,14 @@ public TestResultAction(TestResult result, BuildListener listener) {
         this((Run) null, result, listener);
     }
 
+    @SuppressWarnings("deprecation")
     @Override public Collection<? extends Action> getProjectActions() {
-		return Collections.<Action>singleton(new TestResultProjectAction(run.getParent()));
+        Job<?,?> job = run.getParent();
+        if (/* getAction(Class) produces a StackOverflowError */!Util.filter(job.getActions(), TestResultProjectAction.class).isEmpty()) {
```
oleg-nenashev: Hopefully nobody adds this action as a transient one

src/main/java/hudson/tasks/junit/TestResultAction.java
jglick: The call to `Job.getActions()` cannot be avoided here. Normally it should not be used, but this is one of the exceptions.

src/main/java/hudson/tasks/junit/TestResultAction.java
jglick: If it comes from `JobProperty.getJobActions` (as in `XUnitPublisher`), or is added by `TransientProjectActionFactory`, it will be in `Job.getActions()` and thus we will not add a second copy—that is the purpose of this PR.

oleg-nenashev:
:+1:

src/main/java/hudson/tasks/junit/TestResultAction.java
oleg-nenashev: agreed

PR #19, opened 2015-05-07
[WiP] Remove exception from being thrown on empty/non-existent test result

joshshinn:
In certain scenarios, such as customized Maven builds where specific modules are skipped, an exception would be thrown due to an empty test result file.  Now, no exception is thrown and the build will be able to continue.  See https://issues.jenkins-ci.org/browse/JENKINS-25657

@reviewbybees

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitParser.java
+++ b/src/main/java/hudson/tasks/junit/JUnitParser.java
@@ -113,7 +113,7 @@ public class JUnitParser extends TestResultParser {
             if (files.length == 0) {
                 // no test result. Most likely a configuration
                 // error or fatal problem
-                throw new AbortException(Messages.JUnitResultArchiver_NoTestReportFound());
+                return new TestResult();
             }
 
             TestResult result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio);
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -150,13 +150,9 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep {
             }
             action.setHealthScaleFactor(getHealthScaleFactor()); // overwrites previous value if appending
 			if (result.isEmpty()) {
-                if (build.getResult() == Result.FAILURE) {
-                    // most likely a build failed before it gets to the test phase.
-                    // don't report confusing error message.
-                    return;
-                }
 			    // most likely a configuration error in the job - e.g. false pattern to match the JUnit result files
-				throw new AbortException(Messages.JUnitResultArchiver_ResultIsEmpty());
+                listener.getLogger().println(Messages.JUnitResultArchiver_ResultIsEmpty());
+                return;
 			}
 
             // TODO: Move into JUnitParser [BUG 3123310]
--- a/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
+++ b/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
@@ -274,5 +274,11 @@ public class JUnitResultArchiverTest {
             }
         }
     }
+    @Test public void emptyDirectory() throws Exception{
+        JUnitResultArchiver a = new JUnitResultArchiver("TEST-*.xml");
+        FreeStyleProject freeStyleProject = j.createFreeStyleProject();
+        freeStyleProject.getPublishersList().add(a);
+        j.assertBuildStatusSuccess(freeStyleProject.scheduleBuild2(0));
+    }
 
 }


jglick:
:+1:

oleg-nenashev:
LGTM, but it breaks the original behavior. Probably, it makes sense to add "Don't fail on missing results" checkbox to the advanced section.

jglick:
I see no reason for a new option. There is nothing in the original behavior to break. At most a build which previously failed will now succeed with a warning.

oleg-nenashev:
xUnit plugin [provides such option][1], so IMO it makes sense to have a consistent behavior.

@jglick 
The may cause issues e.g. when a user distributes e-mails with test reports, etc. So the modified behavior may cause glitches in further publisher steps affecting users.

[1]: https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/lib/dtkit/descriptor/TestTypeDescriptor/config.jelly#L79-82

jglick:
@oleg-nenashev yes you can contrive an example where the _absence_ of test reports is an expected failure condition, but this seems a weird use case to me that does not warrant a new UI option that everyone has to see and think about. Normally if there are no test reports, it is because the build already failed in some build step.

At any rate, if you think this is necessary for compatibility and wish to request that the author add such an option, I will stand aside.

oleg-nenashev:
@jglick 
This case is not so significant for me, because I commonly use xUnit plugin. You're the de-facto plugin's owner, so I don't want to push my vision and make @joshshinn to do extra work. If there's no thumbs down from @jenkinsci/everyone , I'm OK to merge the PR as is.

// Seems the plugin's repo has been created improperly (no "Anyone"), and we have no so may watchers here.

jglick:
Hmm, not sure I wanted to be the owner. :-)

What is the issue with the repository? Maybe I did something wrong when splitting this out of core.

oleg-nenashev:
Off-topic: @jglick, I suppose you've manually created the repo instead of asking ircbot to do it (or probably there was a glitch in the bot). In such case https://github.com/orgs/jenkinsci/teams/everyone team does not get and access to the repo and does not gets subscribed to notifications by default.

aheritier:
Strange I disabled many GH notifications previously (especially on
jenkinsci org) but I'm receiving this notification too and I'm not watching
it.

On Tue, May 12, 2015 at 10:08 PM, Jesse Glick <notifications@github.com>
wrote:

> Hmm, not sure I wanted to be the owner. :-)
>
> What is the issue with the repository? Maybe I did something wrong when
> splitting this out of core.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/junit-plugin/pull/19#issuecomment-101405347>
> .
>



-- 
-----
Arnaud Héritier
http://aheritier.net
Mail/GTalk: aheritier AT gmail DOT com
Twitter/Skype : aheritier

mparfianowicz:
Normally if there are no test reports, it is because the build already failed in some build step.

Indeed, this is the most common scenario. But I have a contrary example: the Clover code coverage tool has a test optimization feature, which runs only a subset of tests - it runs only tests relevant with modified files. It can happen that optimization will determine that there are no tests to be executed. In such case there will be no test reports.

jglick:
> It can happen that optimization will determine that there are no tests to be executed. In such case there will be no test reports.

That is exactly the use case motivating this PR: you do not want the publisher to fail.

jglick:
FWIW https://github.com/jenkinsci/junit-plugin/settings/collaboration does list `@everyone`.

hughperkins:
I would think we would want to make this an option, ideally a backwardly-compatible option?  So, by default it fails, but if pass in 'pass_if_no_tests = true', then it passes even if no tests, for example.

jglick:
Well there was not really consensus. Sounds like adding the option to succeed when there are no tests (presumably in an _Advanced_ block, with the default being the current behavior) is the least controversial way forward.

rbywater:
This possibly isn't possible, but could new installs default to the new way
and updates to the old way? (Or would that just be even more confusing?)

Richard

jglick:
> could new installs default to the new way and updates to the old way?

It is technically possible.

> Or would that just be even more confusing?

I think it would be.

jglick:
> could new installs default to the new way and updates to the old way?

Actually I may have misread this to imply a distinction between new vs. updated versions of the _plugin_, but that is not really necessary. It would suffice to have

```java
private boolean allowEmptyResults;
public boolean isAllowEmptyResults() {return allowEmptyResults;}
@DataBoundSetter public void setAllowEmptyResults(b) {allowEmptyResults = b;}
```

```xml
<f:advanced>
    <f:entry field="allowEmptyResults">
        <f:checkbox default="true"/>
    </f:entry>
</f:advanced>
```

Any existing publisher will have an XStream form omitting this field, so will load as `false`. If you round-trip it, will be explicitly `false`. But if you add the publisher fresh, unless you poke around in the _Advanced_ block, it will be set to `true`.

Similarly, for Workflow calls,

```groovy
step([$class: 'JUnitResultArchiver', testResults: '…'])
```

will keep the old behavior, but the _Snippet Generator_ will by default suggest

```groovy
step([$class: 'JUnitResultArchiver', testResults: '…', allowEmptyResults: true])
```

dcoraboeuf:
Hi,

Will there be any progress around this? Our scenario needs this kind of "does not fail if no test":

We have a pre-build job, which is a Gradle incremental build. In case of some small changes, no new test is run, and the JUnit publisher will fail because of the error mentioned above.

I think this is a valid use case for having an additional `allowEmptyResults` parameter.

tomasbjerre:
+1 for merging as is

mparfianowicz:
dcoraboeuf:

In case you can't wait for this fix, then I suggest using a workaround in your builds. You can add an additional step in your build configuration, in which an empty test xml file (with one fake test name) will be created in case of absence of test results.

mparfianowicz:
@joshshinn: will you add this 'allowEmptyResults' toggle to this pull request?

stephenc:
:bee: I cannot see anything wrong with the change... however I think it would be better if the suggestion around retaining the existing behaviour for existing jobs could be implemented... though I suspect that would require:

```
private Boolean allowEmptyResults;
public Boolean getAllowEmptyResults() {return allowEmptyResults;}
@DataBoundSetter public void setAllowEmptyResults(Boolean b) {allowEmptyResults = b;}
```

with a readResolve method that sets the value to `FALSE` otherwise jobs will get this setting the first time they are saved IIUC

rsandell:
:bee:

jtnord:
I am :bug: on this without it being configurable.

I am aware of several setups where a failure in the code creating the test results is not correctly trapped due to buggy tools - and this will then cause builds that should not succeed to be passed as stable - and promoted to production.

ptomato:
Perhaps this could be merged as-is and those who would like a backwards compatibility setting could then submit a follow-up pull request?

oleg-nenashev:
Marked as WiP according to the comments above

christ66:
@oleg-nenashev I think https://github.com/jenkinsci/junit-plugin/pull/28 makes this change obsolete.

oleg-nenashev:
@christ66 seems so

christ66:
Closing as it was resolved in #28 .

PR #20, opened 2015-05-22
[JENKINS-18696] Fix some tests are not reported

harcher81:
Some tests file were not reported when test-suite name and timestamps are the same. This can happen for example if the user run the same test on 2 platforms.

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResult.java
+++ b/src/main/java/hudson/tasks/junit/TestResult.java
@@ -234,8 +234,7 @@ public final class TestResult extends MetaTabulatedResult {
             // JENKINS-12457: If a testsuite is distributed over multiple files, merge it into a single SuiteResult:
             if(suiteResult.getName().equals(suiteResultToAdd.getName()) && nullSafeEq(suiteResult.getId(), suiteResultToAdd.getId())) {
                 for (CaseResult caseResult: suiteResultToAdd.getCases()) {
-                    if (!suiteResult.caseAlreadyExists(caseResult)) {
-                        suiteResult.addCase(caseResult);
+                    if (suiteResult.addCase(caseResult)) {
                         caseResult.replaceParent(suiteResult);
                         duration += caseResult.getDuration();
                     }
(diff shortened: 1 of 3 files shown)


olivergondza:
Tests and JIRA issue needed.

oleg-nenashev:
@harcher81 
Please reference original PRs if you re-create them.

The current PR is a new attempt after https://github.com/jenkinsci/junit-plugin/pull/15

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -229,28 +229,22 @@ private void parsePossiblyEmpty(File reportFile) throws IOException {
...
-                    return;
-                }
-            
-                for (CaseResult cr: sr.getCases()) {
-                    s.addCase(cr);
-                    cr.replaceParent(s);
+            if(suiteResult.getName().equals(suiteResultToAdd.getName()) && nullSafeEq(suiteResult.getId(), suiteResultToAdd.getId())) {
+                for (CaseResult caseResult: suiteResultToAdd.getCases()) {
+                    if (!suiteResult.caseAlreadyExists(caseResult)) {
+                        suiteResult.addCase(caseResult);
```
oleg-nenashev: IMO <code> caseAlreadyExists</code> should be called within the addCase() method to avoid misusages

harcher81:
Can you merge this PR?

oleg-nenashev:
@harcher81 
There is an open comment from @olivergondza . If he agrees with the merge without tests, we can go forward

olivergondza:
I find it quite severe Jenkins can generate false positive (blue) results when it does not report some test results. I have to stick with my -1. Btw, Jesse made the same comment on the previous incarnation of the fix.

harcher81:
There is already a hack on the JUnit plugin, see https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/junit/TestResult.java#L235

I think this fix will produce fewer false positive. But I 'm not against simply remove the current hack without replacement verification.

khneal:
FYI - I submitted another PR (#120) to fix the same issue based on the current master branch. If that one is approved/merged, then this one can be closed.

khneal:
Update - I closed #120 in preference for #131, which I think is a better solution.

PR #21, opened 2015-07-13
[FIXED JENKINS-29383] Fix handling of special chars in aggregated matrix reports

olivergondza:
The path can contain url-encoded special characters. Passing it to the javascript
function over `href="javascript:..."` decodes at least some of the characters so
`showFailureSummary` function search decoded version of element IDs and fails when
the decoded string differs. This change uses onclick event instead, where arguments
are not url-decoded.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,5 +48,15 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!--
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
:+1:

PR #23, opened 2015-07-24
[FIXED JENKINS-29613] Failure summary jumps to the top of the page when expanded/collapsed

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-29613


oleg-nenashev:
@olivergondza 
Thanks! I was not sure that I can provide a valuable feedback on WebUIs, so I've finally forgotten to comment at all

PR #24, opened 2015-08-21
Make it easier for scripts to retrieve passed and skipped test results

aroskuski:
Currently, it is very simple to get a list of failed tests with a single line of code by calling `getFailedTests()` on the `TestResultAction`.  However, getting a list of passed or skipped tests is significantly more difficult.  To do that, you have to dig down to the `PackageResult` level, which not involves traversing further down into the data structures, but also means that you have to collect the results from each package individually.

The changes in this pull request aim to allow passed and skipped test results to be retrieved just as easily as failed ones by changing the way passed and skipped results are handled internally to be more similar to how failed results are handled.

### DIFF
--- a/src/main/java/hudson/tasks/junit/PackageResult.java
+++ b/src/main/java/hudson/tasks/junit/PackageResult.java
@@ -201,7 +201,7 @@ public final class PackageResult extends MetaTabulatedResult implements Comparab
      * @return the children of this test result, if any, or an empty collection
      */
     @Override
-    public Collection<? extends hudson.tasks.test.TestResult> getPassedTests() {
+    public List<CaseResult> getPassedTests() {
         List<CaseResult> r = new ArrayList<CaseResult>();
         for (ClassResult clr : classes.values()) {
             for (CaseResult cr : clr.getChildren()) {
@@ -220,7 +220,7 @@ public final class PackageResult extends MetaTabulatedResult implements Comparab
      * @return the children of this test result, if any, or an empty list
      */
     @Override
-    public Collection<? extends TestResult> getSkippedTests() {
+    public List<CaseResult> getSkippedTests() {
         List<CaseResult> r = new ArrayList<CaseResult>();
         for (ClassResult clr : classes.values()) {
             for (CaseResult cr : clr.getChildren()) {
--- a/src/main/java/hudson/tasks/test/AbstractTestResultAction.java
+++ b/src/main/java/hudson/tasks/test/AbstractTestResultAction.java
@@ -260,7 +260,9 @@ public abstract class AbstractTestResultAction<T extends AbstractTestResultActio
      * A shortcut for scripting
      * 
      * @return List of passed tests from associated test result.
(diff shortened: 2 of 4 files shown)


src/main/java/hudson/tasks/test/AbstractTestResultAction.java
```diff
@@ -257,6 +257,24 @@ public TestResult findCorrespondingResult(String id) {
     }
 
     /**
+     * A shortcut for scripting
+     * 
+     * @return List of passed tests from associated test result.
+     */
```
oleg-nenashev: Add <code>@since TODO</code> and NonNull annotations

oleg-nenashev:
From my POV the change is reasonable, because it follows the "failedTests" pattern. 

The only **serious** concern is related to the impact on the performance and on the memory size. If we have a big test suite (e.g. 10k tests) in the continuous integration flow, we usually expect several tests to fail due to the recent change, so the number of saved test results is minimal. BTW, saving of passed tests in such frequent builds may bloat the RAM memory.

I would propose to make the behavior alterable somehow. E.g. you could keep the original counters for getXXXCount() and record arrays only if such mode is enabled in TestResult and PackageResult constructors.

aroskuski:
How about I make it so that the lists of passed and and skipped test results are only generated on demand when `getPassedTests()` or `getSkippedTests()` are called?  That way, they only get generated if the user intends to use them.

oleg-nenashev:
Yes, such approach should work.

aroskuski:
Okay, all your feedback should be addressed.

I'm not 100% sure that I did `@Nonnull` correctly, however.  I've only seen it used on function arguments before.

oleg-nenashev:
Nonnull is OK.

I think that now we can get into concurrency issues, so it makes sense to make new getter methods synchronized.

aroskuski:
How's this?

oleg-nenashev:
@aroskuski Thanks for fixing the comments. Hopefully I'll test the change and integrate it by the next weekend

src/main/java/hudson/tasks/junit/PackageResult.java
```diff
@@ -201,7 +201,7 @@ public boolean hasChildren() {
      * @return the children of this test result, if any, or an empty collection
      */
     @Override
-    public Collection<? extends hudson.tasks.test.TestResult> getPassedTests() {
+    public List<CaseResult> getPassedTests() {
         List<CaseResult> r = new ArrayList<CaseResult>();
```
olivergondza: Is this change not going to break ABI compatibility for existing callers?

oleg-nenashev: It's an override, so it should be safe

oleg-nenashev:
:+1:

oleg-nenashev:
@olivergondza Ready to go IMHO

oleg-nenashev:
ehh, I've forgotten to squash it :( my bad

PR #25, opened 2015-08-25
Export the TestActions on the TestObject

wolfs:
to make them accessible via the REST api.

### DIFF
--- a/src/main/java/hudson/tasks/test/TestObject.java
+++ b/src/main/java/hudson/tasks/test/TestObject.java
@@ -34,6 +34,7 @@ import jenkins.model.Jenkins;
 
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.*;
+import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 import com.google.common.collect.MapMaker;
@@ -217,6 +218,7 @@ public abstract class TestObject extends hudson.tasks.junit.TestObject {
      * Get a list of all TestActions associated with this TestObject. 
      */
     @Override
+    @Exported(visibility = 3)
     public List<TestAction> getTestActions() {
         AbstractTestResultAction atra = getTestResultAction();
         if ((atra != null) && (atra instanceof TestResultAction)) {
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/tasks/test/TestObject.java
```diff
@@ -217,6 +218,7 @@ public AbstractTestResultAction getTestResultAction() {
      * Get a list of all TestActions associated with this TestObject. 
      */
     @Override
+    @Exported(visibility = 999)
```
daniel-beck: That's a bit much, isn't it?

wolfs: Suggest a value, I will use tree=... anyway.

oleg-nenashev:
It cannot be done in such way => :-1: for the merge

<code>TestAction</code> is not an <code>ExportedBean</code>. If the getter provides a list with a child, which does not has <code>ExportedBean</code> annotation, the API will return a horrible exception instead of the original response. It may lead to serious regressions in plugins depending on JUnit functionality.

There is a similar story in Jenkins core: https://github.com/jenkinsci/jenkins/pull/1698

wolfs:
@oleg-nenashev How does https://github.com/jenkinsci/jenkins/blob/9c443c8d5bafd63fce574f6d0cf400cd8fe1f124/core/src/main/java/hudson/model/Actionable.java#L92-L93 work, then? Action also has no @ExportedBean annotation and this is in jenkins core.

I can add an @ExportedBean annotation to TestAction, too. Would this then be okay for you?

oleg-nenashev:
> https://github.com/jenkinsci/jenkins/blob/9c443c8d5bafd63fce574f6d0cf400cd8fe1f124/core/src/main/java/hudson/model/Actionable.java#L92-L93 work, then?

I' not sure how it works... Probably it's not being exposed by default due to requests depth or uses another Stapler version. It makes sense to evaluate it in any case. I'll do it.

> I can add an @ExportedBean annotation to TestAction, too. Would this then be okay for you?

Yep, it should work well

src/main/java/hudson/tasks/test/TestObject.java
oleg-nenashev: I would just leave the default visibility (3?) if Jenkins shows it in the API in such case.

wolfs:
@oleg-nenashev better now?

oleg-nenashev:
Yes, it's OK for me now :+1: 
I think it makes sense to revise other actions and fields to export the important ones (e.g. getResult()), but it's a topic for another PR

PR #27, opened 2015-09-02
Adding archiveId

rsandell:
Mostly handy in workflows where you might
parse test results in different parts of the flow
and need to separate them somehow in later processing.

Example usage: https://github.com/jenkinsci/parallel-test-executor-plugin/pull/16


rsandell:
Mostly uploading to see if I'm on the right track

@reviewbybees

varmenise:
looks like you would need to use assertEquals(expectedValue, actualValue, delta) in L SuiteResultTest.java:155. Still trying to understand why it did not fail before since it looks like you did not change anything related to that

varmenise:
I think it s something related to the fact that now you re using `org.junit.Assert.*` rather than `junit.framework.Assert.*`

src/test/java/hudson/tasks/junit/SuiteResultTest.java
```diff
@@ -38,6 +38,10 @@
 
 import junit.framework.TestCase;
 
+import static org.hamcrest.core.StringStartsWith.startsWith;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
 /**
```
varmenise: this is causing the [test] (https://jenkins.ci.cloudbees.com/job/plugins/job/junit-plugin/98/org.jenkins-ci.plugins$junit/testReport/hudson.tasks.junit/SuiteResultTest/testSuiteResultPersistence/) to fail , you should modify the test to be aligned with the `org.junit.Assert` API

rsandell: I modified it slightly and amended the commit, to see if it works. The problem is that I don't get the error locally, only on CI.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-suitePrefix.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    A prefix to put on each parsed suite's name.
```
jtnord: an options prefix

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -45,19 +47,30 @@
 public class JUnitParser extends TestResultParser {
 
     private final boolean keepLongStdio;
+    private final String suitePrefix;
```
oleg-nenashev: Add <code>@CheckForNull</code>

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -74,6 +75,12 @@
      */
     private boolean keepLongStdio;
 
+
+    /**
+     * A prefix to put on all test suite names, null == none.
+     */
+    private String suitePrefix = null;
```
oleg-nenashev: doh

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -158,6 +162,9 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio) throws
             String pkg = suite.attributeValue("package");
             if(pkg!=null&& pkg.length()>0)   name=pkg+'.'+name;
         }
+        if (Util.fixEmpty(suitePrefix) != null) {
```
oleg-nenashev: And what about spaces trimming?

jtnord: surely time to refactor to a StringBuilder now :-)

oleg-nenashev: +1

oleg-nenashev:
In general the PR looks good to me. 

Why do you prefer prefixes? WithParams and other modifiers commonly add suffixes instead of prefixes.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -33,6 +33,9 @@ THE SOFTWARE.
 	<f:entry field="keepLongStdio" title="">
 		<f:checkbox name="keepLongStdio" title="${%Retain long standard output/error}"/>
 	</f:entry>
+	<f:entry title="${%Suite name prefix}" field="suitePrefix">
```
oleg-nenashev: I think it should be moved to the Advanced section.

rsandell: There is no advanced section, so I chose to have it open instead of introducing one.

oleg-nenashev: I have no strong opinion on this topic< but for me this option seems to be too narrow specific to be kept in the main config

rsandell:
@oleg-nenashev since the suite name is most often based on ```packageName + className``` adding a suffix would not be as descriptive imho as a prefix, since the suffix is usually the method name.

rsandell:
@varmenise the test is fixed

varmenise:
looks good to me :bee:

stephenc:
adding this info to the suite name smells like a hack vs what you really want to do, i.e. differentiate between different captures of results, so

:bug:

rsandell:
After a talk with @stephenc I'm going to rewrite this into a separate field in the test suite instead of prefixing the name.

rsandell:
Squashed previous commits into the new one, so history is new.

@reviewbybees

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -74,6 +75,13 @@
      */
     private boolean keepLongStdio;
 
+
+    /**
+     * A prefix to put on all test suite names, null == none.
```
stephenc: javadoc seems incorrect

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -71,6 +73,11 @@
     private String timestamp;
     /** Optional ID attribute of a test suite. E.g., Eclipse plug-ins tests always have the name 'tests' but a different id. **/
     private String id;
+    /**
+     * An optional archive id to late be able to differentiate suites that has been merged.
```
stephenc: this sen-tense maketh no cents

stephenc:
seems ok but for the javadoc comments

jtnord:
@rsandell  where's the UI changes?
Surely this needs to be exposed in the web ui for browsing test results - after all whats the point in showing 100 "JenkinsInjectedTests" if you can't distinguish them?

jtnord:
:bug: no corresponding UI changes when browsing the results.

rsandell:
@jtnord test suites aren't rendered in the UI at all, as it is today. Only package and method and this change doesn't affect those.

jtnord:
@rsandell in that case there is still an outstanding bug - that would likely conflict with this change - so :bug:

If when something fails and the user can not tell which test failed in which project from the testResult page then something is broken.  I see the fix for this (<--) and the fix for suitegrouping as different views onto the same thing.

rsandell:
@jtnord something like this? ;)

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -345,7 +365,11 @@ public String getFullName() {
      * @since 1.515
      */
     public String getFullDisplayName() {
-    	return TestNameTransformer.getTransformedName(getFullName());
+        StringBuilder transformedName = new StringBuilder(TestNameTransformer.getTransformedName(getFullName()));
```
stephenc: I would prefer

``` 
return this.archiveId == null ? TestNameTransformer.getTransformedName(getFullName()) : TestNameTransformer.getTransformedName(getFullName()) + "[" + this.archiveId + "]";
```

or an equivalent. We should not be allocating a `StringBuilder` for the verbatim case

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -100,13 +100,35 @@
...
+        SuiteResult suite = testResultAction.getResult().getSuite("hudson.security.HudsonPrivateSecurityRealmTest");
+        assertNotNull(suite);
+        assertEquals("dasId", suite.getArchiveId());
+    }
+
     @RandomlyFails("TimeoutException from basic")
-   @LocalData
-   @Test public void slave() throws Exception {
+    @LocalData
+    @Test public void slave() throws Exception {
```
stephenc: Inherited sin. put the `@Test` annotation on a new line before the `public` modifier

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -100,13 +100,35 @@
 		wc.getPage(build, "testReport/hudson.security"); // package
 		wc.getPage(build, "testReport/hudson.security/HudsonPrivateSecurityRealmTest/"); // class
 		wc.getPage(build, "testReport/hudson.security/HudsonPrivateSecurityRealmTest/testDataCompatibilityWith1_282/"); // method
-		
-
 	}
 
+    @LocalData
+    @Test public void basicSuiteArchiveId() throws Exception {
```
stephenc: Put the `@Test` on a new line before the `public` modifier

stephenc:
Mostly :ant:'s but still you should fix them!

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -106,7 +115,7 @@ private static float parseTime(Element testCase) {
         return 0.0f;
     }
 
-    CaseResult(SuiteResult parent, Element testCase, String testClassName, boolean keepLongStdio) {
+    CaseResult(SuiteResult parent, Element testCase, String testClassName, boolean keepLongStdio, @CheckForNull String archiveId) {
```
tfennelly: Do you need an overloaded version for the original method in case something is extending this? Maybe that's not possible or a concern.

rsandell: The constructor is package access only.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -264,7 +275,11 @@ private static String getSkippedMessage(Element testCase) {
     }
 
     public String getDisplayName() {
-        return TestNameTransformer.getTransformedName(testName);
+        StringBuilder transformedName = new StringBuilder(TestNameTransformer.getTransformedName(testName));
+        if (this.archiveId != null) {
+            transformedName.insert(0, "[" + this.archiveId + "]");
+        }
+        return transformedName.toString();
```
tfennelly: NIT: Seems like a funky way of appending the strings to each other. Isn't this what you're effectively doing...

```java
        StringBuilder transformedName = new StringBuilder();
        if (this.archiveId != null) {
            transformedName.append("[" + this.archiveId + "]");
        }
        transformedName.append(TestNameTransformer.getTransformedName(testName));
```

Seems a little easier to read to me, but obviously subjective (hence NIT).

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -345,7 +365,11 @@ public String getFullName() {
      * @since 1.515
      */
     public String getFullDisplayName() {
-    	return TestNameTransformer.getTransformedName(getFullName());
+        StringBuilder transformedName = new StringBuilder(TestNameTransformer.getTransformedName(getFullName()));
+        if (this.archiveId != null) {
+            transformedName.insert(0, "[" + this.archiveId + "]");
+        }
+        return transformedName.toString();
```
tfennelly: Duplication (of funkiness :) )

stephenc:
:bee:

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -146,26 +156,33 @@ private static void parseSuite(File xmlReport, boolean keepLongStdio, List<Suite
...
+     * @param archiveId
+     *      an optional id from the {@link JUnitResultArchiver} to later differentiate suites when merged.
      */
-    private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio) throws DocumentException, IOException {
-    	this.file = xmlReport.getAbsolutePath();
-        String name = suite.attributeValue("name");
-        if(name==null)
+    private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, @CheckForNull String archiveId) throws DocumentException, IOException {
+        this.archiveId = Util.fixEmptyAndTrim(archiveId);
+        this.file = xmlReport.getAbsolutePath();
```
tfennelly: Absolute path?

tfennelly:
:bee:

src/main/java/hudson/tasks/junit/SuiteResult.java
rsandell: Don't ask me, it was there before I messed with it :)

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -72,6 +72,13 @@
     private transient SuiteResult parent;
 
     private transient ClassResult classResult;
+    /**
+     * An optional archive id to be able to differentiate suites that has been merged.
+     * @see JUnitResultArchiver#archiveId
+     * @see SuiteResult#archiveId
+     */
+    @CheckForNull
+    private final String archiveId;
```
jtnord: this is already available in parent.getArchiveID()

When de-serialized this will no longer be the same string everywhere but a new String - such that if you have 1000 classes you will have 1001 instance of a the same String rather than one instance that is shared - as such it will slow down the loading as well as add more pressure on the GC (lazily loading these does cause issues) - it will also take up (a small amount more diskspace).

Better to just call the parent to get it.
If you feel you really want it locally then make it transient and set it in `freeze()`, `setParentSuiteResult()` `replaceParent()`

src/main/java/hudson/tasks/junit/CaseResult.java
jtnord: there is no need for this ctor signature change as the property is available in the parent.  (but see my above general comment)

jtnord:
:bug: missing `archiveId` in ClassResult and also inefficient propagation of the archiveId.

rsandell:
@jtnord ClassResult is merged together too much to be reliable, it will contain cases from all the splits and so the archiveId in there would only reflect the first case that was added in the merge.

Think of the ATH where all plugin tests are in the plugin package.

stephenc:
You could at least check the string for equality with the parent and copy the parent's ref if `this.artifactId != null && this.artifactId != parent.artifactId && this.artifactId.equals(parent.artifactId)` (could also consider `String.intern()` but I would use a system property to enable interning rather than always intern as there could be a performance impact)

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -345,7 +359,12 @@ public String getFullName() {
      * @since 1.515
      */
     public String getFullDisplayName() {
-    	return TestNameTransformer.getTransformedName(getFullName());
+        String archiveId = this.parent != null ? Util.fixEmptyAndTrim(parent.getArchiveId()) : null;
```
jtnord: calling fixEmpty and trim repeatedly is not the best as the String will be searched for spaces repeatedly.
better to fixEmptyAndTrim when it is set in the parent and just then call parent.getArchiveId()

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -264,7 +264,12 @@ private static String getSkippedMessage(Element testCase) {
     }
 
     public String getDisplayName() {
-        return TestNameTransformer.getTransformedName(testName);
+        String archiveId = this.parent != null ? Util.fixEmptyAndTrim(parent.getArchiveId()) : null;
```
jtnord: calling` fixEmptyAndTrim` repeatedly is not the best as the String will be searched for spaces repeatedly.
better to `fixEmptyAndTrim` when it is set in the parent and just then call `parent.getArchiveId()`

src/main/java/hudson/tasks/junit/ClassResult.java
```diff
@@ -189,6 +200,21 @@ else if(r.isPassed()) {
                 failCount++;
             }
             duration += r.getDuration();
+            if (arId == null) {
```
jtnord: so this looks a little funky to me...
the results are including all cases regardless of the archiveId that the result was from.

src/main/java/hudson/tasks/junit/ClassResult.java
```diff
@@ -219,7 +250,16 @@ public String getFullName() {
     }
     
     public String getFullDisplayName() {
-    	return getParent().getDisplayName() + "." + TestNameTransformer.getTransformedName(className);
+        String arId = Util.fixEmptyAndTrim(this.archiveId);
```
jtnord: no to fixEmptyAndTrim here.

jtnord:
I get the feeling that this is missing something, and there is no clear design outline on exactly what we want to achieve (just adding archiveId would fix the immediate issue - but would it work with matrix style jobs etc and provide the features that a user would require - or would yet another refactoring and migration of persisted data need to be performed causing pain in the near-term future).   Based on this I feel I need to vote :bug:

jglick:
I think the idea is fine but this needs to be part of a larger feature permitting specialized display of results from different stages of a build, and/or on different axis combinations à la `matrix-project`. Until that concept is proven to work, we risk introducing an API that is too weak or otherwise misdesigned, but it might be impossible to enhance compatibly after a release is cut. Can you hold off for a while? I think the downstream PR in `parallel-test-executor` is not exactly a blocker since you can always add extra `excludes` to the final split results, at the cost of bucket partioning being less accurate.

rsandell:
Yes, we are working around the issue by having separate workflows doing the various test phases.

rsandell:
@reviewbybees done 
Closing this since there are talks of a different approach in workflow that this might interfere with.

PR #28, opened 2015-10-10
Added option allowing empty test results.

zen-devconsoft:
This is an attempt at solving one aspect of JENKINS-12815 (https://issues.jenkins-ci.org/browse/JENKINS-12815). This implementation does not allow for setting the build status on missing files, but a simple check-box to allow empty results at all. I hope it is a reasonable compromise between the functionality needed by many (myself included) to allow for missing test files and not to clutter the GUI with too many options.

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitParser.java
+++ b/src/main/java/hudson/tasks/junit/JUnitParser.java
@@ -1,18 +1,18 @@
 /*
  * The MIT License
- * 
+ *
  * Copyright (c) 2009, Yahoo!, Inc.
- * 
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
- * 
+ *
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
- * 
+ *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


zen-devconsoft:
Updates:
I realized I've broken backward compatibility by changing the constructor of one of the semi-internal classes. Updated the class with overloaded constructor with new new parameters instead; preserving the old interface as well.

zen-devconsoft:
It solves JENKINS-22747 (https://issues.jenkins-ci.org/browse/JENKINS-22747), which is a variant of JENKINS-12815.

jplandolt:
Glad that we're on the same page.  When will this pull get accepted?

oleg-nenashev:
I suppose somebody should take an action and to merge pending PRs and release the plugin.
The plugin has no maintainer, so somebody from @jenkinsci/core could do it. Added to my TODO list for the weekend.

jplandolt:
Thanks Oleg

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -39,25 +39,38 @@
...
     public JUnitParser() {
-        this(false);
+        this(false, false);
     }
 
     /**
      * @param keepLongStdio if true, retain a suite's complete stdout/stderr even if this is huge and the suite passed
      * @since 1.358
      */
+    @Deprecated
```
jtnord: nit: newly deprecated code should also have a javadoc `@deprecated` tag saying what you should use instead.

oleg-nenashev:
Sorry, didn't find time to handle this on the weekend. It is in my TODO queue and impacts my karma, so I hope to do it ASAP

jplandolt:
Thanks for letting us know, Oleg. Do you need any support after it's
managed?

jplandolt:
Any sense on when this is going to happen?

oleg-nenashev:
I've performed manual tests on the weekend and everything seems to be fine. Going forward with the merge

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -25,22 +25,28 @@ THE SOFTWARE.
...
-	</f:entry>
-	</j:if>
+    </f:entry>
+    </j:if>
+    <f:advanced>
+        <f:entry title="${%Allow empty results}" field="allowEmptyResults">
+            <f:checkbox default="false" title="${%Do not fail the build on empty test results}"/>
+        </f:entry>
+    </f:advanced>
+    -->
```
rpionke: i think this line should be removed, the plugin version 1.10 got already released today
someextrachars

rpionke:
@oleg-nenashev i think this line "-->" in https://github.com/jenkinsci/junit-plugin/blob/master/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly #51 should be removed
someextrachars

oleg-nenashev:
@rpionke 
Sure

PR #30, opened 2015-10-20
[JENKINS-22747] Add Job Configuration Option to NOT fail JUnit Report publish when no test result files

jplandolt:
Fix for [JENKINS-22747](https://issues.jenkins-ci.org/browse/JENKINS-22747).  Allows the for each job to be configured so as to allow zero JUnit Result files. This is for projects that do not yet have any unit test results.

The default is false; jobs with zero JUnit Result files will fail the job.  This is the current default behaviour of the plugin (as it should be)


jplandolt:
Oops!!

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -230,6 +245,21 @@ public boolean isKeepLongStdio() {
...
+	/**
+	 * @return the ignoreNoReports
+	 */
+	public boolean isIgnoreNoReports() {
+		return ignoreNoReports;
+	}
+
+	/**
+	 * @set the ignoreNoReports
+	 * @since 1.596.2
```
oleg-nenashev: It's not the plugin's version. May be confusing

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -91,25 +97,28 @@ public JUnitResultArchiver(String testResults) {
...
             DescribableList<TestDataPublisher, Descriptor<TestDataPublisher>> testDataPublishers) {
-        this(testResults, false, testDataPublishers);
+        this(testResults, false, false, testDataPublishers);
     }
 	
 	@Deprecated
 	public JUnitResultArchiver(
 			String testResults,
             boolean keepLongStdio,
+            boolean ignoreNoReports,
```
oleg-nenashev: Changes the deprecated API. BTW it's formally a binary compatibility change

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -75,6 +75,12 @@
     private boolean keepLongStdio;
 
     /**
+     * If true, and if no JUnit output, don't fail the JUnit reporting.
+     * @since 1.596.2
```
oleg-nenashev: Wrong version

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
jplandolt: What should the version number be used here, please?

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
jplandolt: What should the version number be used here, please?

oleg-nenashev: The current plugin version is 1.10-SNAPSHOT . Presumably it will be merged into 1.10, so makes sense to add this version. BTW the recommended practice is just to use <code>@since TODO</code> and leave to the person who integrates the PR

oleg-nenashev:
:+1: for the contents.
I'm not sure about the impact of permission changes

zen-devconsoft:
More or less same solution has been offered in pull-request #28, with the difference that the option is placed in "advanced" section and it provide unit tests to prove the functionality.

jplandolt:
Ok good to know!  I'm happy to drop this request as soon as pull-request [#28](http://github.com/jenkinsci/junit-plugin/pull/28) is accepted.

christ66:
#28 has been accepted, so I will mark as closed. Thank you @jplandolt for working on this!

jplandolt:
And it's closed and the branch is in the wind. thanks :)

PR #31, opened 2015-11-11
Fixed test failures on core versions >= 1.597

rsandell:
This change also fixes hpi:run not having the plugin's extensions registered
due to circular dependencies.

Content of the `descriptions` field of `AbstractTestResultAction` caused old builds
not to be fully populated during the conversion so for example the status was kept as FAILURE
instead of SUCCESS when read from disk. As the tests don't want to test that part
it was a simple operation of removing that field from the build.xml so that the tests
had correct data to use.

@reviewbybees


oleg-nenashev:
I would propose to put in on hold till there is a technical need to update the core dependency in order to support new features/fixes

rsandell:
@oleg-nenashev There is a technical need, the PCT can't run this plugin.

jglick:
Can you be more specific about what was wrong with `AbstractTestResultAction.descriptions` and what you changed in the binary files? From a casual reading of the description it sounds like you are suppressing some kind of legitimate problem.

rsandell:
@jglick Yes, I guess I am suppressing some legitimate issue with the deserialization of 'descriptions'. Build records created with later versions of core doesn't have this problem since the field seems to be ignored. The error is swallowed up by core somewhere and not shown, and my debugging efforts couldn't find where or why it happens either.

My change simply removes the descriptions tag from build.xml since they don't contain anything usable.

stephenc:
:bee:

kwhetstone:
:bee:

jglick:
> Build records created with later versions of core doesn't have this problem since the field seems to be ignored. The error is swallowed up by core somewhere

But what about build records created by older versions of core and read by the new version of core? We have to support that, it is a normal user scenario.

rsandell:
IIUC we are talking about build records created by Hudson era build records, so not really a normal user scenario ;)

kwhetstone:
There is a chance that a customer is importing a job from that long ago, but it's pretty rare.  From what I remember thankfully the jobs are imported correctly (as in the states are right), and the user can still see the results of the "most recent" passing test by going to the list of test runs.  Essentially, all the information is there, just not the handy link for the "latest results".  So it's a bug with a workaround.

jglick:
> we are talking about Hudson era build records

OK…what version specifically? Where are you getting this information from? There is no mention of such an age in the PR description.

jglick:
Superseded I think.

PR #32, opened 2015-11-20
remove ambiguity in test errors

paulmillar:
The phrase `<init>` as a test name is used in three places to
represent three independent types of error.  This is needlessly
ambiguous.  This patch removes that ambiguity so it is clear
which case is under consideration.


oleg-nenashev:
weak :+1: for the improvement. BTW it makes sense to avoid formatting changes in unchanged lines

paulmillar:
Sorry about the unrelated trailing white-space changes; different projects have different policies.

paulmillar:
FWIW, the motivation, in part, comes from here:

https://issues.jenkins-ci.org/browse/JENKINS-31673

oleg-nenashev:
:+1:

PR #34, opened 2016-02-01
Convert to parent-pom

kwhetstone:
Converted to parent pom to avoid test failures due to upgraded versions of htmlunit in core.  There are still test failures with junit, but those are unrelated to the parent pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>2.3</version>
     </parent>
     <artifactId>junit</artifactId>
     <version>1.11-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Please do not forget @reviewbybees 
:bee: anyway

pom.xml
```diff
@@ -17,6 +17,10 @@
             <url>http://opensource.org/licenses/MIT</url>
         </license>
     </licenses>
+    <properties>
+        <jenkins.version>1.580.1</jenkins.version>
+        <jenkins-test-harness.version>1.580.1</jenkins-test-harness.version>
```
andresrc: Have you tried compiling against JTH 2.1 (the default)?

jglick: Yeah prefer use of the current test harness.

kwhetstone: If I do that, I have to update the junit test classes to the new htmlunit.  It won't compile if I use 2.1.  I could make those changes here, but I was thinking it would be handled in a different pull request.

andresrc: In other plugins conversions the same pull request has included the,  update the parent POM,  test fixes so that they compile with JTH 2.1 and javadoc fixes so that it does the build does not break on Java 8 If the changes are not too big, encourages adoption.

amuniz: +1 to update JTH

jglick:
Superseded I think.

PR #35, opened 2016-02-19
Testsuite duration may be parsed from time attr

jmarnold1021:
Testsuite duration can be retrieved from its time attribute if
it exists. Otherwise it will continue to be calculated from the sum
of it's testcases. ClassResult duration is now obtained from it's
cases' SuiteResult since it was just re-summing case durations.

### DIFF
--- a/src/main/java/hudson/tasks/junit/ClassResult.java
+++ b/src/main/java/hudson/tasks/junit/ClassResult.java
@@ -188,10 +188,12 @@ public final class ClassResult extends TabulatedResult implements Comparable<Cla
             else {
                 failCount++;
             }
-            duration += r.getDuration();
+            //retrieve the class duration from these cases' suite time
+            if(duration == 0){
+                duration = r.getSuiteResult().getDuration();
+            }
         }
     }
-
 
     void freeze() {
         this.tally();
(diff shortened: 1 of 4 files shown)


olivergondza:
It is useful to describe what tools are using setting this field. Also, it is necessary to provide test case for the new feature.

jmarnold1021:
Hi Oliver, I will look into adding a test case to the pr but I am not sure what tools you are referring to. Do you need to know how the plugin was built to test the new feature? What tools we use the feature for and why the time attribute is necessary?

olivergondza:
I see even junit put the `time` attribute on suite level, that makes my question irrelevant. Sorry

jmarnold1021:
Yea it's mainly because the the JUNIT spec uses it. Also we have been parsing it out an rewriting it into one of the test cases with the others set to 0 in the report.xml in order to get the entire suite time. This feels like a better solution.

jmarnold1021:
Hi again, I have added the test case for the new feature to the pr. Unfortunately we have not been able to receive a jira account to open an official ticket on the issue, our request does not seem to make it through. If the ticket is a necessary requirement we would appreciate a second look into our account request. We really believe this feature to be more intuitive when viewed along side the JUINT Report Specs which would be helpful for all the plungin's users.

jmarnold1021:
Just checking in again, is there any status on whether this pr will be merged or not?

olivergondza:
LGTM

ryanbmilbourne:
@olivergondza @jenkinsadmin when might this PR be merged?  I'd also find this useful.

PR #41, opened 2016-03-18
Update parent POM

jglick:
Allows `mvn hpi:run` to work, via https://github.com/jenkinsci/maven-hpi-plugin/pull/29.

@reviewbybees esp. @andresrc and CC @emanuelez


andresrc:
:bee:

PR #43, opened 2016-04-26
[FIXED JENKINS-34407] Parse time correctly in SuiteResult.

mfriedenhagen:
* #35 introduced getting suite time from attribute.
* However this did not work correctly when time was > 999.

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -36,8 +36,6 @@ import org.kohsuke.stapler.export.Exported;
 
 import java.io.File;
 import java.io.IOException;
-import java.text.DecimalFormat;
-import java.text.ParseException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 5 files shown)


olivergondza:
LGTM

EamonNerbonne:
Note that in some cultures, `,` is the decimal seperator.

mfriedenhagen:
Hello, I know, because in Germany we use the comma as well for this. However, the Jenkins plugin has no chance to guess what localization was used while running the tests. The locales of the JVMs runnng Jenkins, Maven and e.g. maven-surefire-plugin might all be different. There is AFAIK no standard attribute in the junit-xml by which we may determine the locale, until now there is not even a common XSD. surefire is always assuming a English locale for time, see https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd

EamonNerbonne:
Well, if that's the case, I'm not convinced it's better to guess at a value than it is to reject uncertain input.  Removing comma's is going to cause weird corruption in some cultures (especially where period is the thousands separator), so it might be simpler to specify on exact serialization and reject everything non-conforming.

PR #44, opened 2016-05-04
[JENKINS-34598] - Hardening time attribute checks

BernhardBerbuir:
This change is a bug fix for the case when the time attribute of a test suit is empty. This can happen when all tests of a test suite are ignored. Currently an `java.io.IOException` / `java.lang.NumberFormatException` is thrown (see [JENKINS-34598](https://issues.jenkins-ci.org/browse/JENKINS-34598) - JUnit plugin: archiving of test reports fails when time attribute of test suite is empty)

### DIFF
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -165,7 +165,8 @@ public final class SuiteResult implements Serializable {
         this.timestamp = suite.attributeValue("timestamp");
         this.id = suite.attributeValue("id");
         // check for test suite time attribute
-        if( ( this.time = suite.attributeValue("time") ) != null ){
+        this.time = suite.attributeValue("time");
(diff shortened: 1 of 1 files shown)


olivergondza:
I see that your changes are applied atop of https://github.com/jenkinsci/junit-plugin/pull/43/ but is it still a problem in 1.13?

BernhardBerbuir:
Version 1.13 also fixes the problem. Nonetheless applying the pull request might be useful as:
- assigning a variable at a condition is a bad programming practice
- handling a missing time attribute (`time==null`) and an empty time attribute (`time.equals("")`) the same same way would be more consistent

oleg-nenashev:
> Nonetheless applying the pull request might be useful as

I agree, but maybe better to use standard lib methods, which also handle cases with orphaned spaces

timja:
this has been conflicted for a long time, closing, feel free to reopen after resolving conflicts

PR #45, opened 2016-05-13
Prevent wrapping for test name table cell

phoenix384:
The no-wrap class attribute was added some time ago to prevent the duration column from wrapping its text.
But for long test names the test name column still inserts a line break between the plus icon (to show/hide the failure details) and the test name.
This change should fix that.


olivergondza:
I tried to apply this fix in firefox's developer tools and it does not seem to work. To my understanding it prevents _text_ to break but here it is spread between 2 `a` elements.

phoenix384:
The class "no-wrap" just adds the CSS "white-space: nowrap".
I added "white-space: nowrap" to this table cell via Simple Theme Plugin and everything is fine.
It prevents a whitespace from causing a line break.

olivergondza:
Oh, I see it only affects aggregated matrix report. Amending with 596f5de. Thanks.

PR #47, opened 2016-05-31
Add junit Pipeline step and junitWrapper Pipeline step.

abayer:
More testing to be done, needs documentation, needs better error propagation.

- [x] Write the actual step.
- [ ] Help documentation for the global variable.
- [ ] Figure out how to ensure that nested errors still get surfaced even when the `JUnitResultArchiver` `step` fails - right now they're swallowed.

cc @reviewbybees

### DIFF
--- a/src/main/resources/hudson/tasks/junit/pipeline/JUnitPipelineStep.groovy
+++ b/src/main/resources/hudson/tasks/junit/pipeline/JUnitPipelineStep.groovy
@@ -1,0 +1,56 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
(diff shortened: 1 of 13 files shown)


jglick:
See #9.

abayer:
Meaning what? =)

jglick:
Meaning this has been done before, I had some reservations, read up on existing approaches and all associated material.

abayer:
So you're saying not to do it in the junit plugin, even with optional dependencies and extensions?

jglick:
No optional dependencies. If you want to define a step, define it. There is a reason `workflow-step-api` provides no GUI: so there is no cost to depending on it.

abayer:
fwiw, tests pass fine with Java 8, since, well, Permgen.

abayer:
@jglick - So...don't use optional dependencies, meaning don't do it in junit-plugin due to 1.x bundling, meaning do it in a new plugin?

abayer:
Closing this in favor of the `junit-pipeline-steps` plugin that's currently waiting to be moved to the `jenkinsci` org.

jglick:
> don't use optional dependencies, meaning don't do it in junit-plugin due to 1.x bundling, meaning do it in a new plugin?

No, meaning simply add the `workflow-step-api` dependency to the `junit` plugin. Why not?

abayer:
I am so confused. =) I thought that adding the `workflow-step-api` dependency would create problems in 1.x due to bundling?

jglick:
> I thought that adding the `workflow-step-api` dependency would create problems in 1.x due to bundling?

Well, we are not going to produce any more 1.x releases, so nothing you do in a future release of this plugin is going to matter for 1.x bundling, right?

abayer:
Valid point, that. I'll reopen this. =)

src/main/resources/hudson/tasks/junit/pipeline/JUnitWrapperPipelineStep.groovy
```diff
@@ -0,0 +1,56 @@
...
+import org.jenkinsci.plugins.workflow.cps.CpsScript
+
+class JUnitWrapperPipelineStep implements Serializable {
+    CpsScript script
+
+    JUnitWrapperPipelineStep(CpsScript script) {
+        this.script = script
+    }
+
+    def call(Map args, Closure closure) {
```
jglick: Uh, what?? Just make a plain old `Step`.

In fact this DSL is just sugar over the existing `SimpleBuildStep` and so is less valuable than the original PR, which

* allowed you to inspect results in Groovy code and make decisions based on them
* allowed you to decide whether or not to set the build status

src/main/resources/hudson/tasks/junit/pipeline/JUnitPipelineStep.groovy
```diff
@@ -0,0 +1,47 @@
...
+        this.script = script
+    }
+
+    def call(Map args) {
+        String testResults = args.containsKey("testResults") ? args.testResults : ""
+        boolean keepLongStdio = args.containsKey("keepLongStdio") ? args.keepLongStdio : false
+        boolean allowEmptyResults = args.containsKey("allowEmptyResults") ? args.allowEmptyResults : false
+        Double healthScaleFactor = args.containsKey("healthScaleFactor") ? args.healthScaleFactor : 1.0
+
+        script.step($class: "JUnitResultArchiver",
```
jglick: :-1:, this adds no value at all, other than creating a one-off DSL with no `Snippetizer` support that will be obsoleted by JENKINS-29922.

jglick:
Sorry but :-1: on the concept here.

abayer:
Ok, I'll revisit this later, then.

abayer:
Closing - the new metastep syntax suffices.

PR #48, opened 2016-06-13
[JENKINS-28784] Workaround through using raw types for testDataPublishers

abayer:
https://issues.jenkins-ci.org/browse/JENKINS-26535

Required for https://issues.jenkins-ci.org/browse/JENKINS-28784

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -232,7 +232,7 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep {
         this.healthScaleFactor = Math.max(0.0, healthScaleFactor);
     }
 
-    public @Nonnull List<? extends TestDataPublisher> getTestDataPublishers() {
+    public @Nonnull List<TestDataPublisher> getTestDataPublishers() {
         return testDataPublishers == null ? Collections.<TestDataPublisher>emptyList() : testDataPublishers;
     }
 
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -241,7 +241,7 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep {
      *
      * @since 1.2
      */
-    @DataBoundSetter public final void setTestDataPublishers(@Nonnull List<? extends TestDataPublisher> testDataPublishers) {
+    @DataBoundSetter public final void setTestDataPublishers(@Nonnull List<TestDataPublisher> testDataPublishers) {
         this.testDataPublishers = new DescribableList<TestDataPublisher,Descriptor<TestDataPublisher>>(Saveable.NOOP);
         this.testDataPublishers.addAll(testDataPublishers);
     }
--- a/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
+++ b/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
@@ -354,23 +354,19 @@ public class JUnitResultArchiverTest {
         assertTrue(j.getTestDataPublishers().isEmpty());
         assertEquals(args, model.uninstantiate(model.instantiate(args)));
 
-        // Test roundtripping of an instantiated list of test data publishers.
-        args.put("testDataPublishers", Collections.<TestDataPublisher>singletonList(new MockTestDataPublisher("test")));
-        Map<String,Object> explicit = model.uninstantiate(model.instantiate(args));
-        JUnitResultArchiver j2 = model.instantiate(explicit);
-        assertFalse(j2.getTestDataPublishers().isEmpty());
-
-        assertEquals(explicit, model.uninstantiate(model.instantiate(explicit)));
-
         // Test roundtripping from a Pipeline-style describing of the publisher.
-        Map<String,String> describedPublisher = new HashMap<String, String>();
+        Map<String,Object> describedPublisher = new HashMap<String, Object>();
         describedPublisher.put("$class", "MockTestDataPublisher");
         describedPublisher.put("name", "test");
         args.put("testDataPublishers", Collections.singletonList(describedPublisher));
 
         Map<String,Object> described = model.uninstantiate(model.instantiate(args));
-        JUnitResultArchiver j3 = model.instantiate(described);
-        assertFalse(j3.getTestDataPublishers().isEmpty());
+        JUnitResultArchiver j2 = model.instantiate(described);
+        List<TestDataPublisher> testDataPublishers = j2.getTestDataPublishers();
+        assertFalse(testDataPublishers.isEmpty());
+        assertEquals(1, testDataPublishers.size());
+        assertEquals(MockTestDataPublisher.class, testDataPublishers.get(0).getClass());
+        assertEquals("test", ((MockTestDataPublisher)testDataPublishers.get(0)).getName());
 
         assertEquals(described, model.uninstantiate(model.instantiate(described)));
     }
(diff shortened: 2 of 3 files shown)


src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -241,7 +241,7 @@ public double getHealthScaleFactor() {
      *
      * @since 1.2
      */
-    @DataBoundSetter public final void setTestDataPublishers(@Nonnull List<? extends TestDataPublisher> testDataPublishers) {
+    @DataBoundSetter public final void setTestDataPublishers(@Nonnull List<TestDataPublisher> testDataPublishers) {
```
jglick: Not source-compatible, but this seems to have been a pretty low-priority use of wildcards.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -260,7 +260,7 @@ public SimpleArchive(String name, int pass, int fail) {
     @Test public void configRoundTrip() throws Exception {
         JUnitResultArchiver a = new JUnitResultArchiver("TEST-*.xml");
         a.setKeepLongStdio(true);
-        a.setTestDataPublishers(Collections.singletonList(new MockTestDataPublisher("testing")));
+        a.setTestDataPublishers(Collections.singletonList((TestDataPublisher) new MockTestDataPublisher("testing")));
```
jglick: :ant: use

```java
a.setTestDataPublishers(Collections.<TestDataPublisher>singletonList(new MockTestDataPublisher("testing")));
```

abayer: Fixing!

jglick:
OK I suppose, but it would really be better to just fix `DescribableModel` to handle wildcards better. Probably not a very hard fix.

Also in this patch, you have made the getter and setter have different types, which is generally bad.

Would be appropriate to add a `test`-scoped dependency on `structs` and prove that you have fixed the issue.

abayer:
Ok, just add a `test` dependency on `structs`? Nothing else that needs to be done for that? (also, fixing the getter/setter)

jglick:
> just add a `test` dependency on `structs`? Nothing else that needs to be done for that?

Well, add a test using `DescribableModel` that shows you can round-trip one of these things.

abayer:
Ok, I'll see what I can do.

abayer:
I took a look at https://github.com/jenkinsci/pipeline-input-step-plugin/blob/973cd1e816726b2da3d130882336884cbe392434/src/test/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepConfigTest.java#L52-L57 but I can't figure out how that would translate to `new CoreStep(...)`, or what exactly we'd be testing in the first place...hrm.

abayer:
@jglick - any ideas/pointers on that test? Also, ping rest of @reviewbybees - I'd really like to get this in and released so that https://github.com/jenkinsci/junit-attachments-plugin/issues/14 can get merged and released and Pipeline jobs can use attachments. =)

jglick:
[example](https://github.com/jenkinsci/testng-plugin-plugin/pull/28/commits/90bcd062c2a0a7d573591fb713b786fb904885e2#diff-81d80d469a9226bfee96cedb4ef8139eR127)

jglick:
> better to just fix `DescribableModel` to handle wildcards better

Just write a test for it (easy) and then look in [here](https://github.com/jenkinsci/structs-plugin/blob/f22060ee29b8d8c3cf122b1b968448496e7b16a2/plugin/src/main/java/org/jenkinsci/plugins/structs/describable/ParameterType.java#L40-L104) in a debugger.

abayer:
Thanks! I'm just gonna do the test here for now, if that's ok...

Vlatombe:
Hey @abayer, just to let you know I'm also very eager to get junit attachments support on pipeline ;)

abayer:
@jglick - added a round-trip test using `DescribableModel`.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -241,7 +241,7 @@ public double getHealthScaleFactor() {
...
         this.testDataPublishers = new DescribableList<TestDataPublisher,Descriptor<TestDataPublisher>>(Saveable.NOOP);
```
jglick: Hmm, I wonder why this was ever a `DescribableList` to begin with, if (a) we do not allow an `owner` to be set, so you cannot trigger a project save just by editing this list; (b) we do not return ``DescribableList` from the getter, so you cannot call richer methods like `get(Class)`. Anyway not an issue in this PR.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -330,4 +338,40 @@ public String getName() {
...
+        JUnitResultArchiver j = model.instantiate(args);
+        assertEquals("**/TEST-*.xml", j.getTestResults());
+        assertFalse(j.isAllowEmptyResults());
+        assertFalse(j.isKeepLongStdio());
+        assertEquals(1.0, j.getHealthScaleFactor(), 0);
+        assertTrue(j.getTestDataPublishers().isEmpty());
+        assertEquals(args, model.uninstantiate(model.instantiate(args)));
+
+        // Test roundtripping of an instantiated list of test data publishers.
+        args.put("testDataPublishers", Collections.<TestDataPublisher>singletonList(new MockTestDataPublisher("test")));
```
jglick: Of course here the type argument is unnecessary—due to erasure, `DescribableModel` has to make unsafe casts.

jglick: Anyway I would just delete this section. Passing raw `Describable`s as map values is deprecated—supported only for compatible with pre-1.0 Workflow scripts.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -330,4 +338,40 @@ public String getName() {
...
+        Map<String,Object> explicit = model.uninstantiate(model.instantiate(args));
+        JUnitResultArchiver j2 = model.instantiate(explicit);
+        assertFalse(j2.getTestDataPublishers().isEmpty());
+
+        assertEquals(explicit, model.uninstantiate(model.instantiate(explicit)));
+
+        // Test roundtripping from a Pipeline-style describing of the publisher.
+        Map<String,String> describedPublisher = new HashMap<String, String>();
```
jglick: `Object` would be the usual value type, not `String`.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -330,4 +338,40 @@ public String getName() {
...
+        describedPublisher.put("$class", "MockTestDataPublisher");
+        describedPublisher.put("name", "test");
+        args.put("testDataPublishers", Collections.singletonList(describedPublisher));
+
+        Map<String,Object> described = model.uninstantiate(model.instantiate(args));
+        JUnitResultArchiver j3 = model.instantiate(described);
+        assertFalse(j3.getTestDataPublishers().isEmpty());
```
jglick: Best to go ahead and verify that the list has one element, which is a `MockTestDataPublisher`, with the right `getName`.

jglick:
Would still prefer a fix of JENKINS-26535 in `structs`, which I suspect is not hard, but will not insist on it. :bee:

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
abayer: Which section? The roundtripping of an instantiated list?

abayer:
@jglick - I'll take a look at the `structs` stuff eventually, but yeah, not something I can do quickly. =)

jglick:
re:bee:

abayer:
@reviewbybees done

PR #49, opened 2016-06-28
Don't bury AllowEmptyResults in Advanced

joearasin:
I'm not sure if this actually works -- I was taking a 1-minute stab at fixing a usability annoyance. But the 1-item advanced section has caused usability... issues.


jglick:
:+1: I have always wondered why this was not the default in fact. My expectation would be that you get zero or more test results, and if one or more are failing, then the build is unstable. As a diagnostic matter, the publisher can of course issue a _warning_ if it finds no test results, in case you mistyped some file paths or something. I suppose the idea is that it is better to annoy people sometimes than to ever let a build pass that actually had test failures which did not get recorded because of a misconfigured project?

oleg-nenashev:
👍

PR #50, opened 2016-06-29
[JENKINS-36305] Fix TestObject url in order to fix TestActions links

Vlatombe:
@reviewbybees 

https://issues.jenkins-ci.org/browse/JENKINS-36305


jglick:
This seems like it would be incompatible. What was the actual source of the 404? Sounds like a bug in the `junit-attachments` plugin, no? [Here](https://github.com/jenkinsci/junit-attachments-plugin/blob/42807cf1525f42f9b5418bf9268dbf228c385b92/src/main/java/hudson/plugins/junitattachments/AttachmentTestAction.java#L33) specifically.

Vlatombe:
yes indeed it is incompatible. junit-attachments could implement a workaround by building an absolute path at the line you pointed. But I believe the bug lies in junit-plugin, which is using a taglib expecting a different contract on #getUrl().

Vlatombe:
@jglick do you think it is unacceptable to break compatibility here? With the current implementation, **TestAction**s that want to be exposed in sidebar *cannot* use relative urls, unlike regular **Action**s.

jglick:
Perhaps this was not the best design choice, if `junit` foresaw the possibility of test actions being in the sidebar, but changing this behavior now seems too dangerous when there is a simple solution in the plugin actually affected by the bug.

Vlatombe:
I gave a shot to specifying getUrlName with the following contract in junit-attachments
> If the returned string starts with '/', like '/foo', then it's assumed to be
> relative to the context path of the Jenkins webapp.

But then I realised this can't work because of https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/test/TestObject.java#L442. Bottom line: if you don't use relative path, then the action isn't bound so you can't access it.

oleg-nenashev:
@Vlatombe are you still working on it?

Vlatombe:
@oleg-nenashev I still stand by what I said in my last comment. IMO this PR is the only way to fix properly the link in junit-attachment, and I don't see any simple solution to fix it without affecting junit-plugin.

oleg-nenashev:
After some review 👍 from me

src/main/java/hudson/tasks/junit/TestObject.java
```diff
@@ -67,12 +67,14 @@
...
 
-	public abstract String getId(); 	
-	/**
-	 * Returns url relative to TestResult.
+	public abstract String getId();
+
+    /**
+     * Returns the URL of this {@link TestObject}, relative to the context root.
      *
-     * @return the url relative to {@link TestResult}.
```
jglick: Perhaps we could deprecate the old method and introduce a replacement intended to have different behavior?

oleg-nenashev:
> Not sure I understand the compatibility implications well enough to be comfortable merging. Will leave it to @oleg-nenashev to consider.

I should have explained my PoV regarding the compatibility more explicitly. Then I would be able to read it now and restore the context :( Will review from scratch

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -258,14 +258,11 @@ public String getUrl() {
         }
 
         private void generateUrl() {
-            Run<?,?> build = o.getRun();
```
timja, line 261: this code no longer exists on master

timja (commented):
@Vlatombe are you still interested in this PR? or shall we close it.

Vlatombe:
@timja I'll check whether it still makes sense.

Vlatombe:
Updated.

timja:
I’ll take a look next week

src/main/java/hudson/tasks/junit/TestObject.java
jtnord: at first glance I think this will break at least the https://github/com/jenkinsci/cucumber-testresult-plugin

Vlatombe: I had documented the problem from
https://issues.jenkins-ci.org/browse/JENKINS-36305?focusedCommentId=262717&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-262717
I don't see how it can be fixed otherwise.

timja:
I'm struggling to test this any hints?

I have a freestyle build with this:

```text
rm *.xml
wget https://raw.githubusercontent.com/jenkinsci/junit-plugin/master/src/test/resources/hudson/tasks/junit/pipeline/junit-report-testTrends-first-2.xml

mkdir -p org.twia.vendor.VendorManagerTest
echo 'hello' > org.twia.vendor.VendorManagerTest/attachment.txt

echo "[[ATTACHMENT|${WORKSPACE}/org.twia.vendor.VendorManagerTest/attachment.txt]]"
```

and I've configured the publisher =/ 

image

jtnord:
> I'm struggling to test this any hints?
> 
> I have a freestyle build with this:
> 
> ```
> rm *.xml
> wget https://raw.githubusercontent.com/jenkinsci/junit-plugin/master/src/test/resources/hudson/tasks/junit/pipeline/junit-report-testTrends-first-2.xml
> 
> mkdir -p org.twia.vendor.VendorManagerTest
> echo 'hello' > org.twia.vendor.VendorManagerTest/attachment.txt
> 
> echo "[[ATTACHMENT|${WORKSPACE}/org.twia.vendor.VendorManagerTest/attachment.txt]]"
> ```
> 
> and I've configured the publisher =/
> 
> image
> 
> the attachment isn't showing up

normally the "output" needs to be part of the test output not the job console output (and captured by junit (in its sdr-err or std-out element IIRC))

timja (approved):
reproduced the bug and verified that this PR fixes it.

I created https://github.com/timja/junit-attachments-test/blob/master/src/test/java/com/github/AppTest.java#L25 to verify it.

Checking @jtnord 's comment before merging

src/main/java/hudson/tasks/test/TestObject.java
```diff
@@ -104,12 +104,9 @@ public final String getId() {
         return id;
     }
 
-    /**
-     * Returns url relative to TestResult
-     */
     @Override
     public String getUrl() {
-        return '/' + getId();
+        return getRun().getUrl() + getTestResultAction().getUrlName() + "/" + getId();
```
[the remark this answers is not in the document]
timja: not really sure how to test this =/, I tried with the tests in https://github.com/jenkinsci/cucumber-testresult-plugin and these: https://github.com/testdouble/java-cucumber-example

PR #51, opened 2016-06-30
Symbol annotation

kohsuke:
@reviewbybees


jglick:
:bee:

jglick:
(but let us wait for the upstream PR which demonstrates its usage in context)

abayer:
:bee:

PR #52, opened 2016-07-22
optionally configure ignore old tests

johnscancella:
would close JENKINS-5393, JENKINS-9438, and JENKINS-6268
allow for ignoring that tests are old and not failing the build

I am unable to test since the master repo also fails to test. I get an error about class org.w3c.dom.ElementTraversal missing.


factuno-db:
+1 definitely needing this option for testing with Bazel.

PR #53, opened 2016-08-05
[JENKINS-29922] A dependency on symbol-annotation does not work in Jenkins 1.x

jglick:
Or more precisely, prior to 2.2.

@reviewbybees esp. @kohsuke

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,10 +41,9 @@
     </pluginRepositories>
     <dependencies>
         <dependency>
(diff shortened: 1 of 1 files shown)


kohsuke:
:bee:

kohsuke:
Updated https://wiki.jenkins-ci.org/display/JENKINS/Structs+plugin accordingly

jglick:
Thanks, was about to look for the documentation.

The problem was a class loader mismatch, if it was not obvious already.

PR #54, opened 2016-08-22
[JENKINS-37598] Inaccurate aggregation of multiple xmls containing case information about the same testsuite

kgyrtkirk:
[JENKINS-37598](https://issues.jenkins-ci.org/browse/JENKINS-37598)
- [x] change logic to account for all suites involved during time aggregation in `ClassResult`
- [X] moved `TestSuite` merging code from `TestResult` to `TestSuite.merge()`
- [X] replaced `TestSuite.time` field with a boolean
- [X] added 2 testcases - 1 with time attributes; and 1 without them...added to `TestResultTest` (because of `TestSuite.merge()` call location)

### DIFF
--- a/src/main/java/hudson/tasks/junit/ClassResult.java
+++ b/src/main/java/hudson/tasks/junit/ClassResult.java
@@ -33,7 +33,9 @@ import org.kohsuke.stapler.export.Exported;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Cumulative test result of a test class.
@@ -176,7 +178,7 @@ public final class ClassResult extends TabulatedResult implements Comparable<Cla
     @Override
     public void tally() {
         passCount=failCount=skipCount=0;
-        duration=0;
+        Set<SuiteResult>  suites=new HashSet<SuiteResult>();
         for (CaseResult r : cases) {
             r.setClass(this);
             if (r.isSkipped()) {
@@ -188,10 +190,13 @@ public final class ClassResult extends TabulatedResult implements Comparable<Cla
             else {
                 failCount++;
             }
-            //retrieve the class duration from these cases' suite time
-            if(duration == 0){
-                duration = r.getSuiteResult().getDuration();
-            }
+            suites.add( r.getSuiteResult() );
+        }
+
+        // retrieve the class duration from these cases' suite time
+        duration = 0;
+        for (SuiteResult s : suites) {
+            duration += s.getDuration();
         }
     }
 
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -72,8 +72,8 @@ public final class SuiteResult implements Serializable {
     /** Optional ID attribute of a test suite. E.g., Eclipse plug-ins tests always have the name 'tests' but a different id. **/
     private String id;
 
-    /** Optional time attribute of a test suite. E.g., Suites can use their own time attribute or the sum of their cases' times as before.**/
-    private String time;
+    /** Optional true if time attribute is present. E.g., Suites can use their own time attribute or the sum of their cases' times as before.**/
+    private boolean hasTimeAttr;
 
     /**
      * All test cases.
@@ -165,10 +165,11 @@ public final class SuiteResult implements Serializable {
         this.timestamp = suite.attributeValue("timestamp");
         this.id = suite.attributeValue("id");
         // check for test suite time attribute
-        if( ( this.time = suite.attributeValue("time") ) != null ){
-            duration = new TimeToFloat(this.time).parse();
+        this.hasTimeAttr = suite.attributeValue("time")!=null;
+        if( hasTimeAttr ) {
+            duration = new TimeToFloat(suite.attributeValue("time")).parse();
         }
-        
+
         Element ex = suite.element("error");
         if(ex!=null) {
             // according to junit-noframes.xsl l.229, this happens when the test class failed to load
@@ -227,7 +228,7 @@ public final class SuiteResult implements Serializable {
         casesByName().put(cr.getName(), cr);
         
         //if suite time was not specified use sum of the cases' times
-        if(this.time == null){
+        if( !hasTimeAttr ){
             duration += cr.getDuration();
         }
     }
@@ -348,4 +349,22 @@ public final class SuiteResult implements Serializable {
     private static final long serialVersionUID = 1L;
 
     private static final Pattern SUREFIRE_FILENAME = Pattern.compile("TEST-(.+)\\.xml");
+
+    /**
+     * Merges another SuiteResult into this one.
+     * 
+     * @param sr
+     */
+    public void merge(SuiteResult sr) {
+        if (sr.hasTimeAttr ^ hasTimeAttr){
+            throw new IllegalStateException("Merging of suiteresults with incompatible time attribute usage is not supported.( "+getFile()+", "+sr.getFile()+")");
+        }
+        if (hasTimeAttr) {
+            duration += sr.getDuration();
+        }
+        for (CaseResult cr : sr.getCases()) {
+            addCase(cr);
+            cr.replaceParent(this);
+        }
+    }
 }
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -41,6 +41,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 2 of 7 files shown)


src/main/java/hudson/tasks/junit/ClassResult.java
```diff
@@ -188,10 +190,13 @@ else if(r.isPassed()) {
...
-            if(duration == 0){
-                duration = r.getSuiteResult().getDuration();
-            }
+            suites.add( r.getSuiteResult() );
+        }
+
+        // retrieve the class duration from these cases' suite time
+        duration = 0;
+        for (SuiteResult s : suites) {
+            duration += s.getDuration();
```
jglick: Seems like the simpler patch would just be

```diff
-if(duration == 0){
-    duration = r.getSuiteResult().getDuration();
-}
+duration += r.getSuiteResult().getDuration();
```

right?

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -348,4 +349,19 @@ void setParent(hudson.tasks.junit.TestResult parent) {
...
 
     private static final Pattern SUREFIRE_FILENAME = Pattern.compile("TEST-(.+)\\.xml");
+
+    /**
+     * Merges another SuiteResult into this one.
+     * 
+     * @param sr
+     */
+    public void merge(SuiteResult sr) {
+        if (hasTimeAttr) {
```
jglick: Should that be `sr.hasTimeAttr`?

jglick: Would be resolved by test cases checking

```xml
<testsuites name="Automation Tests" tests="2" errors="0" failures="0" ignored="0">
    <testsuite name="test.fs.FileSystemTests">
     <testcase name="testPrefix1" classname="test.fs.FileSystemTest1" time="10"/>
    </testsuite>
    <testsuite name="test.fs.FileSystemTests" time="50">
      <testcase name="testPrefix2" classname="test.fs.FileSystemTest1" time="10"/>
    </testsuite>
</testsuites>
```

and

```xml
<testsuites name="Automation Tests" tests="2" errors="0" failures="0" ignored="0">
    <testsuite name="test.fs.FileSystemTests" time="50">
     <testcase name="testPrefix1" classname="test.fs.FileSystemTest1" time="10"/>
    </testsuite>
    <testsuite name="test.fs.FileSystemTests">
      <testcase name="testPrefix2" classname="test.fs.FileSystemTest1" time="10"/>
    </testsuite>
</testsuites>
```

both of which I would expect to report a total time of 60. (Right?)

src/main/java/hudson/tasks/junit/ClassResult.java
jglick: BTW neither of the test cases seem to check that

```xml
<testsuites name="Automation Tests" tests="2" errors="0" failures="0" ignored="0">
    <testsuite name="test.fs.FileSystemTests">
     <testcase name="testPrefix1" classname="test.fs.FileSystemTest1" time="10"/>
     <testcase name="testPrefix1" classname="test.fs.FileSystemTest2" time="10"/>
    </testsuite>
    <testsuite name="test.fs.FileSystemTests">
      <testcase name="testPrefix2" classname="test.fs.FileSystemTest1" time="10"/>
    </testsuite>
</testsuites>
```

sums to 30.

kgyrtkirk: i'm afraid not...and a change like that would be caught by the one of my new tests
in case not...i should add a case for this too ;)

..the for cycle is running thru cases...and the old trick was to only use the case runtime when the total is 0...the new code collects all the unique suites and aggregates there sum time

src/main/java/hudson/tasks/junit/SuiteResult.java
kgyrtkirk: ...i think we will have time attributest on all testuites or neither of them...i didn't wanted to get lost in the mixed case...but if you think it's important i can fix that ;)

kgyrtkirk: yes that should be hasTimeAttr .. this is the closest currently which rely on the previously mentionet - all of them has it or neither - logic....

kgyrtkirk:
I've updated the testcases - should I squash my changes - instead of adding more commits?

About the mixed cases - i can detect it...but not sure what to do when that happens...throw an exception; or insert a "notice testcase" into the unit results?

        public void merge(SuiteResult sr) {
          if (sr.hasTimeAttr ^ hasTimeAttr){
              // addCase(new CaseResult(this, "__jenkins-junit-plugin_invalid-suite-merge", ""));
              // or throw new RuntimeException()
          }
          ...

oleg-nenashev:
> I've updated the testcases - should I squash my changes - instead of adding more commits?

No need to do it. GitHub allows squashing during the merge

> About the mixed cases - i can detect it...but not sure what to do when that happens...throw an exception; or insert a "notice testcase" into the unit results?

Exception is unlikely a good approach. I would merge with warning

kgyrtkirk:
@oleg-nenashev  i've updated the changes; and included an exception for the mixed case (plus a test case for this mixed case too)

cheers

kgyrtkirk:
@oleg-nenashev @jglick  ping ? months have passed - without progress...1.19 have been rolled out without this fix (about a month ago it seems)...take a look at this report page...sum of testcase times is much less than the package time...it's because of the bug fixed in this pr:
https://builds.apache.org/job/PreCommit-HIVE-Build/lastSuccessfulBuild/testReport/org.apache.hadoop.hive.cli/

I wouldn't want to recommend to update to a non-release version so...please someone could take a look?

kgyrtkirk:
@oleg-nenashev: Can you please take a look at the updated changes?

kgyrtkirk:
@oleg-nenashev can you please take a look at this change?

oleg-nenashev:
@kgyrtkirk sorry, I was snowed under the stuff. Will try to review today.
Anyway, CC @jenkinsci/code-reviewers since more feedback is required

kgyrtkirk:
@jglick  @oleg-nenashev - i've fixed the merge-conflict.

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -72,8 +72,8 @@
     /** Optional ID attribute of a test suite. E.g., Eclipse plug-ins tests always have the name 'tests' but a different id. **/
     private String id;
 
-    /** Optional time attribute of a test suite. E.g., Suites can use their own time attribute or the sum of their cases' times as before.**/
-    private String time;
```
oleg-nenashev: If the class is being persisted on the disk, it requires a migration logic in readResolve()

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -348,4 +349,22 @@ void setParent(hudson.tasks.junit.TestResult parent) {
     private static final long serialVersionUID = 1L;
 
     private static final Pattern SUREFIRE_FILENAME = Pattern.compile("TEST-(.+)\\.xml");
+
+    /**
+     * Merges another SuiteResult into this one.
+     * 
+     * @param sr
```
oleg-nenashev: If you define the parameter in Javadoc, please add its description. Otherwise Javadoc linters will grumble

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -348,4 +349,22 @@ void setParent(hudson.tasks.junit.TestResult parent) {
...
+     */
+    public void merge(SuiteResult sr) {
+        if (sr.hasTimeAttr ^ hasTimeAttr){
+            throw new IllegalStateException("Merging of suiteresults with incompatible time attribute usage is not supported.( "+getFile()+", "+sr.getFile()+")");
```
oleg-nenashev: Since you introduce a new method, it would be preferable to propagate a non-RunTime exception and to make the code above handle it. If you want to stick to `IllegalStateException `, it should be documented at least

src/main/java/hudson/tasks/junit/SuiteResult.java
daniel-beck: Removal of this field will probably result in old data warnings. Should be made deprecated and transient instead.

kgyrtkirk: Oh...I wasn't aware of this; in this case I think the best would be to live with existing 'time' field; and use it

src/main/java/hudson/tasks/junit/SuiteResult.java
kgyrtkirk: I've changed this exception into a logger call...since the severity of this is not really critical - and theres no proper way to handle this elsewhere...so it will log a warning in this case.

kgyrtkirk:
@oleg-nenashev could you please take a look at this? I think the latest changes will look more promising :)

kgyrtkirk:
@oleg-nenashev , @jglick, @jenkinsci/code-reviewers could you please take a look at this patch?

kgyrtkirk:
Could someone merge this? I would like to use this if possible from the next release (and drop the usage of a custom built plugin)

PR #55, opened 2016-08-29
[JENKINS-32148] use alternate background for table rows, highlight row on hover

dirk-thomas:
This patch uses the `stripped` class for tables to give alternate rows a slightly different background color. Additionally it highlights the current row on hover.

Both effects are also used in other parts of the Jenkins UI, e.g. the build history panel on the left side.

This addresses https://issues.jenkins-ci.org/browse/JENKINS-32148

### DIFF
--- a/src/main/resources/hudson/tasks/junit/CaseResult/list.jelly
+++ b/src/main/resources/hudson/tasks/junit/CaseResult/list.jelly
@@ -28,7 +28,7 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define"
          xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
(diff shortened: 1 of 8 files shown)


olivergondza:
How about putting the `hover` functionality intro core together with `stripped` class.

dirk-thomas:
Good point. Once jenkinsci/jenkins#2533 gets merged I will update this to only add the `stripped` classes.

olivergondza:
I am fine to make this stripped right away without the hover functionality that would be picked from core once users upgrade.

dirk-thomas:
Sounds good to me too. I updated the PR to only add the `stripped` class.

olivergondza:
Thanks

dirk-thomas:
Just out of curiosity: when are you usually releasing new versions of the plugin? For now I have deployed a custom built package but it would be great if others could get this enhancement easily.

olivergondza:
The plugin is not actively maintained. I will release this at Monday unless someone does it sooner. :wink:

dirk-thomas:
Thank you!

PR #56, opened 2016-09-30
Added '<' and '>' as unsafe URI characters.

FrelliBB:
I opened this pull request due to the `safe()` function not replacing the `<` and `>` characters. This was triggered due to the `cucumber-testresult-plugin` using the `getSafeName()` for Scenarios and Features as file names when publishing reports (https://github.com/jenkinsci/cucumber-testresult-plugin/blob/master/src/main/java/org/jenkinsci/plugins/cucumber/jsontestsupport/CucumberTestResultArchiver.java).

Cucumber Scenarios can have the `<` and `>` characters in the scenario names (https://github.com/cucumber/cucumber/wiki/Scenario-Outlines) and as such these unsafe characters were preventing the results from being published to Jenkins.

### DIFF
--- a/src/main/java/hudson/tasks/test/TestObject.java
+++ b/src/main/java/hudson/tasks/test/TestObject.java
@@ -399,7 +399,7 @@ public abstract class TestObject extends hudson.tasks.junit.TestObject {
             return "(empty)";
         } else {
             // this still seems to be a bit faster than a single replace with regexp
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Looks good to me 👍 
CC @jtnord , who is a maintainer of cucumber-testresult-plugin

PR #57, opened 2016-10-14
Fix typo: resultadod -> resultados

bryant1410:
Also GH warned: "We've detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8."


olivergondza:
Java properties files are supposed to stay in Latin1 - please do not change that.

bryant1410:
Ok, GH did it automatically. Will revert it later

bryant1410:
Done

PR #58, opened 2016-11-03
CaseResult.java

kavyaAirbus:
Added description about the test result


olivergondza:
Hm, what is this? What is the purpose?

kavyaAirbus:
I Need to Display the Detail description of the test result.As of now i am getting only Status like Passed or failed.I Need to include the link in the test result page where it navigates to page which contains more Detail Information.I thought by changing in the source code it may help me.Correct me if i am wrong.Also if you could know how to do, could you please help me on this.

Thanks

olivergondza:
The change as proposed, have no effect whatsoever as it created new file in location where it is not even complied. Also, from the branch name `patch-1`, I suspect this was not tested in any way.

In order to get this improvement merged (or even reviewed), modify the existing file instead of creating new one, make sure the tests are still passing, add new tests to cover the new functionality if applicable and then propose for merging.

kavyaAirbus:
I am not able to compile it.May i know in which i Need to compile the code?

olivergondza:
This should provide sufficient guidance: https://wiki.jenkins-ci.org/display/JENKINS/Plugin+tutorial#Plugintutorial-BuildingaPlugin

jglick:
Not sure what this is but it does not look like a mergeable fix.

PR #59, opened 2016-11-28
JENKINS-29649-related: AbstractTestResultAction may inadvertently pick action from different plugin

anenviousguest:
When a job uses 2 or more publishers which depend on "junit-plugin" API (e.g. JUnit plugin + TAP plugin), [this check](https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/test/TestResultProjectAction.java#L93) may cause situation when test result actions from one plugin are used to plot graph for another plugin.
 
This was causing [JENKINS-29649](https://issues.jenkins-ci.org/browse/JENKINS-29649), but for that specific one we managed to find work-around on TAP plugin side.

As outlined by @oleg-nenashev in the following discussion thread, one may want to fix this on JUnit plugin side:
* https://github.com/jenkinsci/tap-plugin/pull/14#issuecomment-249346957
* https://github.com/jenkinsci/tap-plugin/pull/14#issuecomment-249355315

Would appreciate you could have a look on that.
Thanks.

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitTestResultProjectAction.java
+++ b/src/main/java/hudson/tasks/junit/JUnitTestResultProjectAction.java
@@ -1,0 +1,24 @@
+package hudson.tasks.junit;
+
+import hudson.model.AbstractProject;
+import hudson.model.Job;
+import hudson.tasks.test.TestResultProjectAction;
+
+
+public class JUnitTestResultProjectAction extends TestResultProjectAction<TestResultAction> {
+
+    public JUnitTestResultProjectAction(Job<?, ?> job) {
+        super(job);
+    }
+
+    @Deprecated
+    public JUnitTestResultProjectAction(AbstractProject<?, ?> project) {
+        super(project);
+    }
+
+
+    @Override
+    public Class<TestResultAction> getTestResultActionClass() {
+        return TestResultAction.class;
+    }
+}
--- a/src/main/java/hudson/tasks/test/TestResultProjectAction.java
+++ b/src/main/java/hudson/tasks/test/TestResultProjectAction.java
@@ -47,7 +47,7 @@ import java.util.List;
  *
  * @author Kohsuke Kawaguchi
  */
-public class TestResultProjectAction implements Action {
+public class TestResultProjectAction<T extends AbstractTestResultAction> implements Action {
     /**
      * Project that owns this action.
      * @since 1.2-beta-1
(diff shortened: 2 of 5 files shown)


anenviousguest:
@oleg-nenashev : could this make sense?
Thanks!

anenviousguest:
did anyone have a chance to look on this? 👀 
Thanks!

anenviousguest:
any chance to get this reviewed? Thanks guys.

oleg-nenashev:
Sorry, missed previous pings. Please use @jenkinsci/code-reviewers if you need more reviewers (at least there will be more people to blame for the missing reviews)

src/main/java/hudson/tasks/test/TestResultProjectAction.java
```diff
@@ -102,6 +102,12 @@ public AbstractTestResultAction getLastTestResultAction() {
         return null;
     }
 
+    @SuppressWarnings("unchecked")
+    public Class<T> getTestResultActionClass() {
```
oleg-nenashev: Would be nice to add some documentation for this method. I'm pretty sure all classes in other plugins are supposed to override this method in order to have the correct behavior

src/main/resources/hudson/tasks/test/TestResultProjectAction/jobMain.jelly
```diff
@@ -26,7 +26,7 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:t="/lib/hudson" xmlns:test="/lib/hudson/test">
     <table style="margin-top: 1em; margin-left: 1em;">
-        <j:set var="tr" value="${it.job.lastCompletedBuild.getAction(it.class.classLoader.loadClass('hudson.tasks.test.AbstractTestResultAction'))}"/>
+        <j:set var="tr" value="${it.job.lastCompletedBuild.getAction(it.testResultActionClass)}"/>
```
oleg-nenashev: Would it be a breaking change if it just starts using `getClass()`?

anenviousguest: @oleg-nenashev i still think it's fine, as we're looking not for `it` instances (which is test result *project* action) but for linked test result actions.

anenviousguest:
@oleg-nenashev thanks for review! Documentation added. Any other thoughts?

src/main/java/hudson/tasks/junit/JUnitTestResultProjectAction.java
```diff
@@ -0,0 +1,24 @@
...
+
+
+public class JUnitTestResultProjectAction extends TestResultProjectAction<TestResultAction> {
+
+    public JUnitTestResultProjectAction(Job<?, ?> job) {
+        super(job);
+    }
+
+    @Deprecated
+    public JUnitTestResultProjectAction(AbstractProject<?, ?> project) {
```
jglick: How can you deprecate a constructor you just added? This makes no sense.

src/main/java/hudson/tasks/test/TestResultProjectAction.java
```diff
@@ -103,6 +103,32 @@ public AbstractTestResultAction getLastTestResultAction() {
     }
 
     /**
+     * <p>
+     * Plugins implementing junit-plugin API should override this method to indicate test result action type to look
```
jglick: Is there a downstream PR showing this being overridden successfully?

timja:
Closing due to inactivity and unresolved issues for over 3 years, please feel free to re-open if you intend to keep working on it.

PR #62, opened 2017-01-16
Allow to print Failure and Error messages together

KKoukiou:
It happens when running tests with pytest that we can have a
test failing with both failure and error.

ex:
import pytest
import logging

@pytest.fixture()
def setup_class(request):
    def finalize():
        raise Exception("test teadown Exception")
    request.addfinalizer(finalize)

def test_1(setup_class):
    assert 1==2

In this case the XML report will contain two nodes, a failure
and an error and both messages are usefull in the jenkins TestReport.
Currently, the plugin chooses to display the error report over
the failure. This patch allows failure message to appear as well.

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -68,6 +68,8 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     private final String skippedMessage;
     private final String errorStackTrace;
     private final String errorDetails;
+    private final String failureStackTrace;
+    private final String failureDetails;
     @SuppressFBWarnings(value = "SE_TRANSIENT_FIELD_NOT_RESTORED", justification = "Specific method to restore it")
     private transient SuiteResult parent;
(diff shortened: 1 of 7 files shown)


rsandell:
please add tests

KKoukiou:
@rsandell I added one test, if you have some better idea how to test this patch please let me know!

olivergondza:
I am not sure this is a way to go. Is this something that JUnit of other XUnit frameworks can generate or it is pytests specific? I am concerned this kind of duality is not expected by clients of API that is being changed here.

KKoukiou:
@olivergondza Thanks for the reply. I am not aware if you can generate it through these frameworks, the way I generate such XML report, is with pytest with --junitxml option.
However, I noticed that the xunit xml xsd does not allow both error and failure for the same testcase,
since it applies <xs:choice minOccurs="0" maxOccurs="unbounded"> for error, failure etc attributes.
So, I understand my patch does not comply with the format and I 'll abandon it.
Thanks for the feedback.

PR #63, opened 2017-01-25
Fix JENKINS-41134

slide:
This adds a test for JENKINS-41134, fix which will make the test pass will come after.

### DIFF
--- a/src/test/resources/hudson/tasks/junit/JENKINS-41134/TestSuite_second.xml
+++ b/src/test/resources/hudson/tasks/junit/JENKINS-41134/TestSuite_second.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+    <testsuite name="TestSuite_second" tests="1" errors="0" failures="1" skip="0" timestamp="2012-04-05T10:50:00">
+        <testcase classname="TestFoo2" name="bar2" time="123.980">
+            <failure message="PROGRAM.cbl:2 Use a program name that matches the source file name" type="WARNING">
+                WARNING: Use a program name that matches the source file name
+                Category: COBOL Code Review – Naming Conventions
+                File: /project/PROGRAM.cbl
+                Line: 2
+            </failure>
+        </testcase>
+    </testsuite>
+</testsuites>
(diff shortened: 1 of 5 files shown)


src/test/resources/hudson/tasks/junit/JENKINS-41134/TestSuite_second.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+    <testsuite name="TestSuite_second" tests="1" errors="0" failures="1" skip="0" timestamp="2012-04-05T10:50:00">
```
jtnord: So the interesting thing with the Jenkins results is that the test suite/names are identical - so allthough this is a good test, it's not reproducing the issue as observed

slide: True, I will add another test. I did verify on an actual build of Jenkins with the Jenkinsfile that the fix does resolve that issue, but a completely valid test is a good idea.

slide: Are we worried about the case when the same exact test runs and has the same exact timestamp? If the two tests with the same name have a different timestamp, they are considered a different instance of the test. If they have the same name, and the same timestamp they are considered the same. What is the likelihood that the same test suite will run at the same exact time on two nodes?

jtnord: my gut says that the probability is low (even though the the timestamp has only second granularity).
At least for the reported failure (as provisioning the windows node takes a while).

Now for something running tests in a parallel branch on the same node (e.g. with different browsers or some flag) then the probability is probably higher - but I think that can be handled as a different issue - as it may break use cases where a user runs unit tests, and then integration (maven failsafe) on the same workspace and runs the junit step in between...

jtnord:
Thanks for looking into this Alex.

src/test/resources/hudson/tasks/junit/JENKINS-41134/TestSuite_second.xml
jamesdh: This exact scenario you described above appears to have bitten us. We have mocha tests that run extraordinarily fast, and sure enough we have three different test suites with the same name, same timestamp, but one of them has failures. Jenkins was only occasionally reporting the failures. Guessing it was just the luck of the draw as to which one it picked to represent the pass/fail state. 

Is there an issue created to represent this scenario?

PR #64, opened 2017-02-03
[JENKINS-41710] Optionally annotate test results with a test run name

stephenc:
See [JENKINS-41710](https://issues.jenkins-ci.org/browse/JENKINS-41710)

Here is the pipeline syntax screen with this change:

screen shot 2017-02-03 at 15 50 14

The test run name does not make sense for non-pipeline jobs, so they do not see it:

screen shot 2017-02-03 at 15 50 33

When you have a pipeline with a testRunName assigned you see results like so:

screen shot 2017-02-03 at 15 45 05

If you mix `junit` steps with and without the `testRunName` parameter, you can get something like this:

screen shot 2017-02-03 at 15 53 23

(names in bold because these are new results from the previous run)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -95,7 +95,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         return new TimeToFloat(time).parse();
     }
 
-    CaseResult(SuiteResult parent, Element testCase, String testClassName, boolean keepLongStdio) {
+    CaseResult(SuiteResult parent, Element testCase, String testClassName, boolean keepLongStdio, String testRunName) {
         // schema for JUnit report XML format is not available in Ant,
         // so I don't know for sure what means what.
         // reports in http://www.nabble.com/difference-in-junit-publisher-and-ant-junitreport-tf4308604.html#a12265700
@@ -119,7 +119,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         }
 
         className = testClassName;
-        testName = nameAttr;
+        testName = testRunName == null ? nameAttr : nameAttr + " (" + testRunName + ")";
         errorStackTrace = getError(testCase);
         errorDetails = getErrorMessage(testCase);
         this.parent = parent;
@@ -200,9 +200,22 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
      * @param testName Test name.
      * @param errorStackTrace Error stack trace.
      */
+    @Deprecated
     public CaseResult(SuiteResult parent, String testName, String errorStackTrace) {
+        this(parent, testName, errorStackTrace, null);
+    }
+    
+    /**
+     * Used to create a fake failure, when Hudson fails to load data from XML files.
+     *
+     * @param parent Parent result.
+     * @param testName Test name.
+     * @param errorStackTrace Error stack trace.
+     * @since FIXME
+     */
+    public CaseResult(SuiteResult parent, String testName, String errorStackTrace, String testRunName) {
         this.className = parent == null ? "unnamed" : parent.getName();
-        this.testName = testName;
+        this.testName = testRunName == null ? testName : testName + " (" + testRunName + ")";
         this.errorStackTrace = errorStackTrace;
         this.errorDetails = "";
         this.parent = parent;
@@ -212,7 +225,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         this.skipped = false;
         this.skippedMessage = null;
     }
-    
+
     public ClassResult getParent() {
     	return classResult;
     }
--- a/src/main/java/hudson/tasks/junit/JUnitParser.java
+++ b/src/main/java/hudson/tasks/junit/JUnitParser.java
@@ -46,6 +46,7 @@ public class JUnitParser extends TestResultParser {
 
     private final boolean keepLongStdio;
     private final boolean allowEmptyResults;
+    private String testRunName;
 
     /** TODO TestResultParser.all does not seem to ever be called so why must this be an Extension? */
     @Deprecated
@@ -71,6 +72,18 @@ public class JUnitParser extends TestResultParser {
     public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults) {
         this.keepLongStdio = keepLongStdio;
         this.allowEmptyResults = allowEmptyResults;
+    }
+
+    /**
+     * @param keepLongStdio if true, retain a suite's complete stdout/stderr even if this is huge and the suite passed
+     * @param allowEmptyResults if true, empty results are allowed
+     * @param testRunName if non-{@code null}, the name of the test run to append to case results.
+     * @since FIXME
+     */
+    public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults, String testRunName) {
+        this.keepLongStdio = keepLongStdio;
+        this.allowEmptyResults = allowEmptyResults;
+        this.testRunName = testRunName;
     }
 
     @Override
@@ -101,7 +114,8 @@ public class JUnitParser extends TestResultParser {
         // also get code that deals with testDataPublishers from JUnitResultArchiver.perform
 
         return workspace.act(new ParseResultCallable(testResultLocations, buildTime,
-                                                     timeOnMaster, keepLongStdio, allowEmptyResults));
+                                                     timeOnMaster, keepLongStdio, allowEmptyResults,
+                testRunName));
     }
 
     private static final class ParseResultCallable extends MasterToSlaveFileCallable<TestResult> {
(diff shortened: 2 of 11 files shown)


yotamshapira:
This actually may make sense in a non-pipeline scenario. Consider the following:
Tests take a long time to run (similar to the scenario specified in the ticket), so we split them across multiple jobs that we run in parallel.
When all tests finish, we copy the tests results to the upstream job and publish the tests results in the aggregated context.
The problem is that all results are shown as a flat list and we don't have a way of differentiating between the results.

cyrille-leclerc:
@stephenc in a maven multi module project, would it make sense to use a name based on the module "groupId:artifactId"?

stephenc:
@cyrille-leclerc well I see the need for a second solution in this space too, along the lines of #27 as that would allow segmenting the results at the start.

cyrille-leclerc:
@stephenc you captured what I have in mind: I would like to have on the Pipeline Job Run page one section per maven module ("groupId:artifactId:version") so that all the reports (Junit, FindBugs...) would be grouped by section.

Note that some Maven executions (e.g. "mvn release:prepare release:perform") will generate reports several time for each Maven module and this is the reason why I want to have the "version" reported in the section name.

stephenc:
@cyrille-leclerc I would see multiple executions of the same tests being annotated with this feature. Separation of test runs being achieved by #27

jtnord:
Note: I would not want to see test results from different maven modules using the approach in #27 as then you would have a mess when you wanted to separate results based on "mvn test" and "ath.sh" and other layers of testing that are build outside of maven but use junit style reporting.

either that or the maven module approach would need to do something special.  But in reality - do we need segregated unit tests per maven module.  Apart from auto generated unit tests that don't use the module groupid:artifactid in the test name (hpi:injectedTests) I have never come across a name collision before.

jglick:
@svanoort needs to review this, because IIRC he had a rather different proposal involving attaching test results directly to `FlowNode`s (rather than the `Run`), which of course would require visualizations such as Blue Ocean to call a new API.

cyrille-leclerc:
@jglick would this `FlowNode` integration you are referring to be able to segregate test results that are generate by multiple modules of 1 unique Maven multi module project built through 1 single `sh` call?

```groovy
docker.image("cloudbees/java-build-tools:2.0.0").inside {
    stage ("Build") {
        withMaven(mavenLocalRepo: '.m2') {
            git "https://github.com/cyrille-leclerc/multi-module-maven-project.git"
            sh "mvn clean install"
        }
    }
}
```

image

jglick:
@cyrille-leclerc no, since all the results would all be on one `FlowNode`. But I agree with @jtnord that this is not a very useful goal anyway.

cyrille-leclerc:
@jglick I am not sure to understand. You mean that it would not be really useful to separate the junit results by artifactId in the case of a Maven multi module project? 

Maven does separate the results of each module and the "Jenkins Maven Project job type" does the same, I think there is a lot of value in this.

jglick:
To be clear, this PR is much closer to my original vision of how differentiated build results should work: it is straightforward enough and flexible. But @svanoort had rather different plans in JENKINS-27395, I suppose with the goal of allowing Blue Ocean to display test results associated with a particular stage or step. So I was asking for him, or someone from the B.O. team, to comment on what in this proposal would be insufficient. Even if this ad-hoc labeling _is_ fine for B.O., there should be a matching downstream PR in `blueocean-plugin` demonstrating consumption of the new API in that UI.

> the "Jenkins Maven Project job type" does the same

I know, and find it useless, but perhaps there are people who really rely on this for some reason.

svanoort:
@jglick Actually I think my proposal is kind of orthogonal to this -- you can track to the flowgraph level where results came from, and still split by test run name, or vice versa. 

I like 'em both to be honest.  They're accomplishing different objectives.

philmcardle:
Not to distract from the conversation over Maven, but I have a similar problem with the output of tests from GoogleTest run against a C++ project that this PR would help with - I have as many tests with the same name as I have combinations of operating system and build config, and this would let me group them clearly.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -200,9 +200,22 @@ private static int halfMaxSize(Collection<CaseResult> results) {
...
+    
+    /**
+     * Used to create a fake failure, when Hudson fails to load data from XML files.
+     *
+     * @param parent Parent result.
+     * @param testName Test name.
+     * @param errorStackTrace Error stack trace.
+     * @since FIXME
+     */
+    public CaseResult(SuiteResult parent, String testName, String errorStackTrace, String testRunName) {
```
svanoort: 🐜  Make use of CheckForNull and NonNull here.   Javadocs leave something to be desired.  Like content.

jglick:
> you can track to the flowgraph level where results came from, and still split by test run name

Doing both seems like needless complexity. We should pick what we need and stick with it.

> I have as many tests with the same name as I have combinations of operating system and build config

Yes, that is the usual use case for something like this, even for Maven (whether single-module or reactor).

BTW this proposal really needs a downstream PR in `parallel-test-executor-plugin` which would allow you to split tests per label. That was in fact the motivation for #27 (cf. https://github.com/jenkinsci/parallel-test-executor-plugin/pull/16).

svanoort:
Jesse I'm not sure where you're coming from, but this and the proposal I had are accomplishing completely different goals.  This lets you create multiple test result for a run with fixed names (useful).  My proposal lets you track test results to where in a flow they were created. 

I would encourage you go and reread https://issues.jenkins-ci.org/browse/JENKINS-27395 - my final proposal includes the nodeId of the flownode the step ran in (which I still think has value) but it attaches it to the Run itself, just to add extra info.

jglick:
> My proposal lets you track test results to where in a flow they were created.

I got that. I am asking what user feature is enabled by that which is not covered just as well by this proposal. We do not want to have to go back in a few months and add yet another field to test results which has partial overlap with an existing one.

cobexer:
I'm a heavy user of matrix jobs for test execution and the changes proposed here don't cut it at all.
@stephenc Did you have a look at what a matrix job does with junit test results?
Image of junit grouping with a matrix job

IMO there are two things that need to be done to make junit(and other test) reporting useful for pipelines where multiple platforms and/or testsuites are involved:

 * a) Same testsuite on multiple platforms: Grouping like in the image I posted here - not by monkey patching the name of the tests.
 * b) Different testsuites (maybe on different platforms too): Multiple reports like the one in the image I posted above.

For a) I think that a map with strings would do the job. Like: { platform: "win32", jre: "7", language: "de_AT" } turning that into a string of the form "key=value;..." would probably make this relatively easy and very powerful. If the map was stored this would even allow to add dynamic ad hoc grouping to be added later.
For b) I think that a simple Tesuite Name parameter would suffice

jglick:
FWIW my original quasi-proposal (never thoroughly developed) was to associate a `Map<String,String>` with each reporting step. The keys could be like matrix axes—thus a given build might contain reports labelled (in the notation of `hudson.matrix.Combination.toString()`):

* `platform=linux,java=7`
* `platform=linux,java=8`
* `platform=windows,java=7`
* `platform=windows,java=8`

or they could be test suite names

* `suite=unit`
* `suite=acceptance`

or some combination

* `suite=unit`
* `suite=acceptance,platform=linux,java=7`
* `suite=acceptance,platform=linux,java=8`
* `suite=acceptance,platform=windows,java=7`
* `suite=acceptance,platform=windows,java=8`

Whatever the keys, it is important that any UI comparing a test result to that of a previous build (“failing since build #1234”) refer to a matching combination. (So, `FlowNode.id` does not make for a good key.) Ideally visualizations could also automatically detect that two or more keys form a space and display grids of results (such as the acceptance test results in the third example above).

Getting this right is IMO fairly important because we would expect the same system to be applied to any other plugin which displays reports—code coverage, static analysis, etc.—that might plausibly need to be grouped.

svanoort:
@jglick My feel is that this change moves the bar forward for users in a meaningful way -- we could argue endlessly about the "best" way to accomplish this, but may be letting perfect become the enemy of good.

If we want to do multivariate tagging, my feel is that the only appropriate way is with TagsActions attached to enclosing blocks, and a mechanism to collect *all* TagsActions applying to a given node.  

Probably something like `FlowScanningUtils.findEnclosingBlocks(thisNode).filter(FlowScanningUtils.hasActionPredicate(TagsAction.class))`

stephenc:
@cobexer this is not intended to be the complete solution, only a fix for one aspect of the problem. FTR I do not see other partitioning schemes as conflicting with this, rather I see them as complementary.

With pipeline, I may not even be running the different runs on different nodes

cyrille-leclerc:
Hi, any progress here?

cyrille-leclerc:
Any reason to block this improvement? cc @svanoort  @jglick

cyrille-leclerc:
ping

jglick:
It needs a coherent design, including Blue Ocean support. Long under discussion with @abayer & @svanoort.

gena01:
There's "hostname" attribute that's in the Junit XML that can be added per testsuite element.  It's also something I might need soon and is something I wish JUnit would support.

cobexer:
@gena01 I downvoted your comment because hostnames are not universally useful.
With dynamic hosts and jenkins node labels the actual hostname is not good, especially if you are interested in things like which build introduced the test error since there you need to be able to correlate the test across multiple builds which might have happened on different hosts.

While hostname may be good enough for lots of people it should not be forced upon everyone.

gena01:
@cobexer I understand your concerns, but i was coming from a Junit XML schema (which somehow is not uniform). and it contains hostname attribute. i am not sure you can "label" things right now. My concern is to be able to build/create something that might be used as an input format for other tools besides Jenkins as well.

jglick:
Any `hostname` attribute in JUnit XML format is irrelevant here because we are concerned with defining groupings at the Jenkins level.

jglick:
@abayer I presume this is superseded by #76 and should be closed?

stephenc:
I'm giving up on this PR

PR #65, opened 2017-03-16
Do not register XMLEntityResolver as a global extension

jerrinot:
The resolver is useful only for parsing TestNG results. There
is no need to register the resolver as a global extension.

Please see a related discussion and reasoning at https://github.com/jenkinsci/jenkins/pull/2806

### DIFF
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -121,7 +121,7 @@ public final class SuiteResult implements Serializable {
 
         // parse into DOM
         SAXReader saxReader = new SAXReader();
-        ParserConfigurator.applyConfiguration(saxReader,new SuiteResultParserConfigurationContext(xmlReport));
+        saxReader.setEntityResolver(new XMLEntityResolver());
 
         Document result = saxReader.read(xmlReport);
         Element root = result.getRootElement();
--- a/src/main/java/hudson/tasks/junit/XMLEntityResolver.java
+++ b/src/main/java/hudson/tasks/junit/XMLEntityResolver.java
@@ -23,10 +23,6 @@
  */
 package hudson.tasks.junit;
 
-import hudson.Extension;
-import hudson.tasks.junit.SuiteResult.SuiteResultParserConfigurationContext;
-import hudson.util.io.ParserConfigurator;
-import org.dom4j.io.SAXReader;
 import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
@@ -48,8 +44,7 @@ import java.util.logging.Logger;
  *
  * @author Mikael Carneholm
  */
-@Extension
-public class XMLEntityResolver extends ParserConfigurator implements EntityResolver {
+public class XMLEntityResolver implements EntityResolver {
 
     private static final String TESTNG_NAMESPACE = "http://testng.org/";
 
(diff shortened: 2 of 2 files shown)


jerrinot:
hi,

any chance to have this PR reviewed?
Thanks!

oleg-nenashev:
It is hard to evaluate the impact of this change. JUnit plugin used to be a part of the Jenkins core, hence there may be various barely traceable dependencies on the original functionality. I would recommend running Jenkins ATH and PCT test suites before merging this change

jerrinot:
@oleg-nenashev: thanks for you feedback, very much appreciated. who can run these test suites?

olivergondza:
I would say that if some plugin depends on that to be set globally, it would start failing as soon as junit is no longer always installed (IOW, Jenkins 2.0). What can be tricky to detect is such expectation by a plugin that depends on junit plugin.

oleg-nenashev:
> I would say that if some plugin depends on that to be set globally, it would start failing as soon as junit is no longer always installed (IOW, Jenkins 2.0)

It was possible to disable it even before 2.0. The plugin has been decoupled around .509.x IIRC. But Jenkins retains all the implicit dependencies via the `detachedPlugins` engine, hence Jenkins won't startup if somebody deletes the JUnit plugin && there is a transient dependency via an old core dependency in a plugin. But I agree it is hard to trace the usages.

> thanks for your feedback, very much appreciated. who can run these test suites?

Good question. @olivergondza and  @andresrc  could probably suggest the best ways to do so without rebuilding the world

jerrinot:
If this change it too risky then I can resurrect the [original PR](https://github.com/jenkinsci/jenkins/pull/2806)
perhaps with some maximum time-to-live. 

The current situation is really suboptimal to put it nicely - there is a RPC for every single XML being processed.

olivergondza:
To my understanding this is triggered iff it is a test case referring to TestNG DTD, which I presume does not happen often. To verify this is safe we need 2 test cases to get such testcase.xml parsed by both the plugins successfully, verify it would fail without the global resolver (the feature is completely untested) and verify it gets back fixed once the "local" resolver is hooked back as this PR suggests.

The only plugin specifically interested in TestNG results seems to be testng-plugin so the impact should not be that hard to verify.

oleg-nenashev:
Since there is not that much prgress with testing, do we want to go forward and release this change as release-candidate to make it testable in the wild?

oleg-nenashev:
But I agree the autotests requested by @olivergondza is a right precondition for the merge

jerrinot:
From what I can see the extension is meant to prevent HTTP call to fetch TestNG DTD when parsing TestNG output files. 

So as long as the machine running the parser has an access to `http://testng.org/testng-1.0.dtd` it will work even when the extension is not registered at all -> that makes it hard to write a test. 

One would need to intercept remote calls I guess?

jerrinot:
This PR is effectively reverting the code to the same state as it was when the resolver was [originally introduced](https://github.com/jerrinot/junit-plugin/commit/50869a1742b887e134a3a684d4c1dbf90ae6b9a9) back in 2007. It was only in 2011 when the concept was generalized and the resolver was [turn into an Extension](https://github.com/jerrinot/junit-plugin/commit/764040e9a1e1789ae939dbb23626dbc08c9f0877).

olivergondza:
I see the testing is a pain here and this clearly is an improvement. To my understanding it can negatively impact only instances that are resolving that entity in different context and can not reach the testng website.

oleg-nenashev:
@rsandell was going to take a look, so I have conditionally assigned him to the ticket

jerrinot:
hello @rsandell, did you have a chance to have a look / test it?

varju:
Hi there.  Sorry to drop in uninvited, but I landed here based on the discussion in https://github.com/jenkinsci/jenkins/pull/2806.

From the comments above, it sounds like this change is approved by all.  Is there anything blocking the merge?

jerrinot:
many thanks to all reviewers!
is there anything preventing to merge this changeset?

abayer:
Nope, there isn't. =) @oleg-nenashev Want me to merge and do a beta release?

oleg-nenashev:
@abayer I am not in the maintainer list, but you are there ;)
https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-junit.yml

jerrinot:
@abayer, @oleg-nenashev: guys, who can merge it? :)

jerrinot:
:fireworks: thanks!

PR #68, opened 2017-04-23
JENKINS-41281 Move BlueJUnitTestResult to JUnit plugin

i386:
We've implemented a small API that provides data to Blue Ocean's REST API for test reporting and restructured `blueocean-rest` so that it is only an API for our REST model.

We needed increase the minimum java version to 7 from 6 for this plugin as the min Java version for Blue Ocean is 7.

The following dependencies have been added to junit plugin for this change:
```
[INFO] +- io.jenkins.blueocean:blueocean-rest:jar:1.1.0-beta-2-SNAPSHOT:compile
[INFO] |  +- io.jenkins.blueocean:blueocean-commons:jar:1.1.0-beta-2-SNAPSHOT:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-databind:jar:2.8.7:compile
[INFO] |  |     +- com.fasterxml.jackson.core:jackson-annotations:jar:2.8.0:compile
[INFO] |  |     \- com.fasterxml.jackson.core:jackson-core:jar:2.8.7:compile
[INFO] |  \- io.jenkins.blueocean.rest.annotation:capability-annotation:jar:1.2:compile
```

PTAL @vivek @jglick @michaelneale

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
     <url>http://wiki.jenkins-ci.org/display/JENKINS/JUnit+Plugin</url>
     <properties>
         <jenkins.version>1.580.1</jenkins.version>
-        <java.level>6</java.level>
+        <java.level>7</java.level>
     </properties>
     <licenses>
         <license>
@@ -45,6 +45,11 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
             <version>1.2</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins.blueocean</groupId>
+            <artifactId>blueocean-rest</artifactId>
+            <version>1.1.0-beta-2-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>io.jenkins.blueocean</groupId>
             <artifactId>blueocean-rest</artifactId>
-            <version>1.1.0-beta-2-SNAPSHOT</version>
+            <version>1.1.0-beta-2</version>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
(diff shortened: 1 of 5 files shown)


olivergondza:
Since the code is minimal and the junit is one of the most used plugins, would this not be better places in blueocean? This adds dependencies for junit, rises Java version, etc. What is the gain?

pom.xml
```diff
@@ -14,7 +14,7 @@
     <url>http://wiki.jenkins-ci.org/display/JENKINS/JUnit+Plugin</url>
     <properties>
         <jenkins.version>1.580.1</jenkins.version>
-        <java.level>6</java.level>
+        <java.level>7</java.level>
```
oleg-nenashev: Without Jenkins core dependency bump this is incorrect 🐛

pom.xml
```diff
@@ -47,6 +47,11 @@
             <version>1.2</version>
         </dependency>
         <dependency>
+            <groupId>io.jenkins.blueocean</groupId>
+            <artifactId>blueocean-rest</artifactId>
+            <version>1.1.0-beta-2-SNAPSHOT</version>
```
oleg-nenashev: It may blow up at any moment since the plugin depends on higher Core version. So it may be missing some API

i386:
@olivergondza @oleg-nenashev the benefit here is that we remove a hard dependency on JUnit for Blue Ocean and the JUnit related code in the ecosystem - including its Blue Ocean implementation - lives in the JUnit plugin. We've trimmed the dependencies so that the dependency on `blueocean-rest` brings in very few dependencies. 

As for bumping Java versions, I am not sure what the resistance is here - core is now 1.8 minimum and AFAIK there are no supported LTS releases that support a 1.6 minimum.

olivergondza:
I do not feel strongly enough to block this.

i386:
Ping :)

oleg-nenashev:
@i386 Ping what? My comments regarding Java version and transient core dependency have not been addressed, so I keep 🐛

i386:
Had a good chat with @oleg-nenashev about this change last night. Unless we bump the core minimum version to 2.7.1 (Blue Ocean's minimum) we can't release a junit plugin with this change.

PR #69, opened 2017-05-10
Fix a file descriptor leak

dgageot:
Dom4j leaks a fd when SaxReader.read(File) is used.
Let's rather use SaxReader.read(FileInputStream)
and close the stream ourself.

This file descriptor leak is specially annoying
on a Windows slave because it prevents from 
deleting the xml report at the end of the build

Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -33,6 +33,7 @@ import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


dgageot:
@oleg-nenashev sorry, no. Should I create one?

olivergondza:
Link to the dom4j would be more helpful.

oleg-nenashev:
@dgageot It's up to the plugin maintainer. I commonly require JIRA issues for bugs, but it is not my plugin. If @olivergondza is fine leaving it without JIRA issue, I am fine as well

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -123,10 +124,19 @@
         SAXReader saxReader = new SAXReader();
         ParserConfigurator.applyConfiguration(saxReader,new SuiteResultParserConfigurationContext(xmlReport));
 
-        Document result = saxReader.read(xmlReport);
-        Element root = result.getRootElement();
+        FileInputStream xmlReportStream = null;
+        try {
+            xmlReportStream = new FileInputStream(xmlReport);
```
olivergondza: Since the `try` block is here for the stream closing only, the instantiation can be moved in front of the `try` block making the reference non-null simplifying the `finally` block as well.

slide:
Is this just waiting on the fix that @olivergondza requested?

src/main/java/hudson/tasks/junit/SuiteResult.java
jglick: Simpler still in `-source 7`, if we just update to 1.625.x or later:

```java
try (InputStream xmlReportStream = new FileInputStream(xmlReport)) {
    parseSuite(xmlReport, keepLongStdio, r, saxReader.read(xmlReportStream).getRootElement());
}
```

PR #71, opened 2017-06-09
Expose test data publicly

Greybird:
This will allow dependent plugins (like xunit one) to run TestDataPublishers

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -177,12 +177,11 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep {
             }
 
             // TODO: Move into JUnitParser [BUG 3123310]
-            List<Data> data = action.getData();
             if (testDataPublishers != null) {
                 for (TestDataPublisher tdp : testDataPublishers) {
                     Data d = tdp.contributeTestData(build, workspace, launcher, listener, result);
                     if (d != null) {
-                        data.add(d);
+                        action.addData(d);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


jglick:
It is good style to also file a downstream PR with a snapshot dependency demonstrating how you would expect to use this new API in another plugin. I `deploy`ed this change as `1.21-20170612.204643-1` which you can use in a POM dependency version, allowing the CI builder for the downstream plugin to run integration tests.

Greybird:
Hi @jglick,

Thanks for your answer and for publishing this version. Will publish the (not compiling) dependency PR next time to match this good practice.
I created the xunit plugin PR jenkinsci/xunit-plugin#48 with the adequate POM dependency.

Greybird:
Hi @oleg-nenashev 

Thanks for your feedback!

Wouldn't set the collection to unmodifyableList kind of defeat the purpose which is to feed/modify data into this collection? (cf https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java#L185)

If we are to create a new method, maybe I can suggest switching from exposing the collection to adding an *addTestData* public method ?
This would solve the plugin usage search issue, keep a somewhat consistent api in terms of naming & duplication, and still allow other plugins to contribute to test data from the outside?

If future developments show a need to actually modify already added data, then we could rediscuss about exposing the collection, 

What do you think about this change in proposal?

Noted about javadoc!

Regards,

Arnaud

oleg-nenashev:
> Wouldn't set the collection to unmodifyableList kind of defeat the purpose which is to feed/modify data into this collection?

It will, I should have read the PR description carefully

`addTestData()` is fine for me. Just make sure the API documentation explains the `save()` behavior (if the method saves atuomatically)

Greybird:
Hi @oleg-nenashev , @jglick ,

As discussed, here is a new proposal.

Regards,

Greybird:
Here is the update.

Greybird:
Hi,

@oleg-nenashev , @jglick , sorry to ask again, but could you review this PR following changes, and see if it can be merged, and the plugin released with this change (and probably others, as I see a few approved PR in the waiting list?)

Arnaud

Greybird:
Thanks a lot!

Greybird:
Hi @oleg-nenashev , @jglick , is this possible to have this PR merged and to release a version of the junit plugin?
I see other PRs approved too, maybe a good opportunity to merge all validated ones at the same time?

PR #72, opened 2017-06-09
Job to aggregate can contain variable, expand it before use

ahahxof:
It seems no existing Unit tests in the project uses env variable so I skip this part. let me know if you want me to add one.


src/main/java/hudson/tasks/test/AggregatedTestResultPublisher.java
```diff
@@ -27,12 +27,14 @@
...
 import hudson.model.AbstractProject;
 import hudson.model.AutoCompletionCandidates;
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
 import hudson.Util;
 import static hudson.Util.fixNull;
 import hudson.model.Action;
 import hudson.model.BuildListener;
+import hudson.model.DependencyGraph;
```
ahahxof: I don't need this, sorry. I will remove it and the excess new lines added below

ahahxof:
@abayer I've added the test. please review

PR #73, opened 2017-07-14
Showing content, of unparsable file, in error message

tomasbjerre:
When using this plugin I got an error with the following message:

```
Failed to read test report file /opt/cm/jenkins/workspace/app-branch-11.0.0-SNAPSHOT/proj/target/surefire-reports/TEST-TestSuite.xml
org.dom4j.DocumentException: Error on line 39323 of document file:///opt/cm/jenkins/workspace/app-branch-11.0.0-SNAPSHOT/proj/target/surefire-reports/TEST-TestSuite.xml : XML document structures must start and end within the same entity. Nested exception: XML document structures must start and end within the same entity.
	at org.dom4j.io.SAXReader.read(SAXReader.java:482)
	at org.dom4j.io.SAXReader.read(SAXReader.java:264)
	at hudson.tasks.junit.SuiteResult.parse(SuiteResult.java:126)
	at hudson.tasks.junit.TestResult.parse(TestResult.java:301)
```

With this commit I would save a lot of time when trying to figure out why.

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResult.java
+++ b/src/main/java/hudson/tasks/junit/TestResult.java
@@ -23,6 +23,9 @@
  */
 package hudson.tasks.junit;
 
+import static com.google.common.io.Files.toByteArray;
+import static hudson.remoting.HexDump.toHex;
+import static java.util.logging.Level.SEVERE;
 import hudson.AbortException;
 import hudson.Util;
 import hudson.model.Run;
@@ -36,6 +39,7 @@ import hudson.tasks.test.TestObject;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.RandomAccessFile;
 import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -44,6 +48,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.logging.Logger;
 
 import org.apache.tools.ant.DirectoryScanner;
 import org.dom4j.DocumentException;
@@ -52,6 +57,7 @@ import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.export.Exported;
 
 import javax.annotation.Nonnull;
+import com.google.common.annotations.VisibleForTesting;
 
 /**
  * Root of all the test results for one build.
@@ -59,6 +65,9 @@ import javax.annotation.Nonnull;
  * @author Kohsuke Kawaguchi
  */
 public final class TestResult extends MetaTabulatedResult {
+    private static final Logger LOGGER = Logger.getLogger(TestResult.class.getName());
+
+    static final int MAX_NBR_BYTES = 100;
 
     /**
      * List of all {@link SuiteResult}s in this test.
@@ -144,7 +153,7 @@ public final class TestResult extends MetaTabulatedResult {
     }
 
     public TestObject getParent() {
-    	return parent;
+        return parent;
     }
 
     @Override
@@ -154,7 +163,7 @@ public final class TestResult extends MetaTabulatedResult {
 
     @Override
     public TestResult getTestResult() {
-    	return this;
+        return this;
     }
 
     @Deprecated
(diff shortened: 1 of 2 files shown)


olivergondza:
I do not like the idea of attaching possibly megabytes of data to Exception message. Especially, when the file has 39323 lines.

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -306,14 +310,20 @@ public void parse(File reportFile) throws IOException {
...
+            try {
+                reportFileContent = Files.toString(reportFile, UTF_8);
+            } catch (Exception ee) {
+                /* No need to further explain the error if we end up here. */
+            }
             if (!reportFile.getPath().endsWith(".xml")) {
-                throw new IOException("Failed to read "+reportFile+"\n"+
-                    "Is this really a JUnit report file? Your configuration must be matching too many files",e);
+                throw new IOException("Failed to parse "+reportFile+"\n"+
+                    "Is this really a JUnit report file? Your configuration may be matching too many files? The file contains:\n\n" + reportFileContent + "\n\n",e);
```
olivergondza: We reject the file because its name does not end with `.xml`, I do not think we should read it.

tomasbjerre: We reject it because it cannot be parsed *and suggest* that it may not be a report file because of its name.

I would argue that this rarely happens. Probably happens during first run of a newly configured job... I don't see a problem with having a very large log in those rare cases.

olivergondza: > We reject it because it cannot be parsed and suggest that it may not be a report file because of its name.

Right, I missed that.

> I don't see a problem with having a very large log in those rare cases.

This is not about build logs (only). You are reading potentially megabytes of data into string eagerly, transferring it over the network and reconstructing it on master side. Putting aside the memory pressure, this is an effective way to dos the system either intentionally or accidentally.

Trimming the file to, say, 1KB would help user to identify non-result files are included but filename would almost just as well IMO. In case of "real" parsing problem trimming can omit the offending part.

The best alternative would be to get SAX report more useful error so we do not need to forward the file to the user at all.

If we absolutely must report the file as a whole, it should be transferred as a file (so compressed and buffered without loading it whole into the memory).

@jenkinsci/code-reviewers, thoughts?

daniel-beck: Full contents shouldn't be transferred. Could be gigabytes of data. Perhaps the first 100 bytes, the last 100 bytes, if ASCII readable with other chars as hex codes (I think the remoting stream diagnosis looks like that and it helps with binary crap), and the total length?

tomasbjerre: I Changed it to include only the last 5 lines.

Not really sure how to replace none readable ascii with hex codes, in a pretty way =) Perhaps there is a utility for that in Jenkins?

daniel-beck: `hudson.remoting.HexDump` might work. Somewhat different purpose but perhaps close enough.

tomasbjerre: Updated PR with HexDump.

daniel-beck:
If this might be binary nonsense, I'd be much more comfortable with a fixed number of bytes (50-100) than "five lines". How does the reverse line reader handle 1 billion `0x00` bytes?

tomasbjerre:
Updated to read only last 100 bytes.

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -41,19 +45,25 @@
...
+
 /**
  * Root of all the test results for one build.
  *
  * @author Kohsuke Kawaguchi
  */
 public final class TestResult extends MetaTabulatedResult {
+    private static final Logger LOGGER = Logger.getLogger(TestResult.class.getName());
+
+	static final int MAX_NBR_BYTES = 100;
```
olivergondza: Please respect indentation in the file.

olivergondza:
@tomasbjerre, looks good to me. Does it still point out the problem you have experienced?

src/main/java/hudson/tasks/junit/TestResult.java
tomasbjerre: fixed

tomasbjerre:
Yes it does! Thanks for all your feedback =)

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -306,20 +316,53 @@ public void parse(File reportFile) throws IOException {
...
                 add(sr);
             }
         }
     }
 
+	@VisibleForTesting
+	String getLastLines(File file, int maxNbrBytes) {
+		RandomAccessFile raf = null;
+		try {
+			byte[] readBytes = null;
```
abayer: Can you please fix the indents to match the code style in general? Thanks.

tomasbjerre:
Indents fixed.

PR #74, opened 2017-07-25
[JENKINS-5393] Add option to allow old test results to be included

ericparton:
This would close JENKINS-5393, JENKINS-9438, and JENKINS-6268 by adding a toggle for the current behavior of ignoring any tests that predate the start of the build. This is based on an earlier pull request by @johnscancella. I made it pipeline friendly and fixed the merge conflict pointed out by @oleg-nenashev.

The new option is disabled by default to preserve backwards compatibility.

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitParser.java
+++ b/src/main/java/hudson/tasks/junit/JUnitParser.java
@@ -46,6 +46,7 @@ public class JUnitParser extends TestResultParser {
 
     private final boolean keepLongStdio;
     private final boolean allowEmptyResults;
+    private final boolean allowOldResults;
 
     /** TODO TestResultParser.all does not seem to ever be called so why must this be an Extension? */
     @Deprecated
@@ -59,8 +60,7 @@ public class JUnitParser extends TestResultParser {
      */
     @Deprecated
     public JUnitParser(boolean keepLongStdio) {
-        this.keepLongStdio = keepLongStdio;
-        this.allowEmptyResults = false;
+        this(keepLongStdio, false, false);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bjfr:
Hi Eric!

I have implemented a fix for this problem also (https://github.com/jenkinsci/junit-plugin/pull/81) unfortunately before seeing your PR :(

Actually I started doing it the same way as you (even used the exact same flag name :), but then I started thinking and realised that the existing flag *allowEmptyResults* actually is enough to fix this. Non existing or too old tests are actually the same, tests which have not run this build.

So I implemented it using the existing flag instead!

We should merge our PR:s into one, maybe we should discuss how to best solve this? Two heads are always better than one! :D

The important thing is to have this fixed, it is a major nuisance which is causing my team unnecessary workarounds to make Jenkins work well with Gradle!

My PR: https://issues.jenkins-ci.org/browse/JENKINS-47315

ericparton:
Hey Björn,

I'm glad to see I'm not the only one trying to use the Gradle incremental builds to speed up CI runs. :)

I took a look at your PR and I think I prefer your solution since it's simpler and less verbose than mine. My only concern with modifying the existing allowEmptyResults flag is that someone could care about the distinction between old or non-existent tests for some odd reason, but that does seem unlikely. 

We should see if we can get one of the maintainers to weigh in your pull request and move it forward.

linux4hach:
I am having a problem with our CI and kind of need this....anybody have any idea when it will be pulled in?

linux4hach:
Ah I see.....just now attempting to build with maven,.....lots of errrors?

abayer:
Can you please resolve the merge conflicts? I'll review this once that's done.

ericparton:
Closed as duplicate of https://github.com/jenkinsci/junit-plugin/pull/108

PR #75, opened 2017-07-26
[JENKINS-27395] Add a junitResults step

abayer:
[JENKINS-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395)

Subsumes #64 

This is still a work-in-progress - I wanted to get the initial work up for review to see if I'm going down the right path. 

- [ ] Tests of `junitResults` step
- [ ] API for pulling all test results for a given name from the `TestResultAction`
- [ ] Probably some way of not displaying the test run name when it's auto-generated - and perhaps a better way to auto-generate than just using the `FlowNode.getId()`
- [ ] Downstream consumption, most notably in Blue Ocean.
- [ ] Consider whether to ditch the `testRunName` stuff at the `CaseResult`/`ClassResult` level and instead drive the distinction through a new field in `SuiteResult` (since `SuiteResult.name` already has enough meaning).

cc @reviewbybees


jglick:
How does this relate to #9?

abayer:
Ha, another attempt at a junit step! I'll see what's reusable from there.

abayer:
@jglick Ok, I'd say this supplants #9 - I prefer the approach I'm taking here of reusing/sharing as much logic and behavior between the step and `JUnitResultArchiver` as possible.

abayer:
Closing in favor of #76

PR #76, opened 2017-07-26
[JENKINS-27395] Add a junitResults step

abayer:
[JENKINS-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395)

I like this one better than #75 (which I'm closing) - it doesn't subsume #64, and doesn't bother with that test run name stuff at all, instead relying on Pipeline itself to get the right information to mark `SuiteResult`s as associated with a particular `Run` and `FlowNode` when appropriate.

This uses the run's externalizable ID and the step's FlowNode.getId()
together as a unique key to track which SuiteResults are associated
with a particular execution, and allows getting a TestResult object
from a run ID and 1 or more node IDs passed to an existing TestResult
object that contains suites for that run/those nodes.

Note that JUnitResultArchiverTest.zip has been moved around and
renamed due to
https://github.com/jenkinsci/jenkins-test-harness/commit/79d9603370d736fa1ede35e78b5f4c421cf2d5be
breaking tests otherwise, since the JUnitResultArchiveTest/ directory
will end up getting picked up as JENKINS_HOME otherwise.

- [x] Full testing of `junit` step
- [x] Snippet Generator config jelly and help
- [x] API for pulling all test results for a run ID/node ID(s) from the `TestResultAction`
- [x] Downstream consumption, most notably in Blue Ocean.
- [ ] Thorough testing of effects of changes to `CaseResult#getDisplayName()`
    * `getFailedSince()` (DONE)
    * Case URL (via `safeName`)
    * ...what else?

cc @reviewbybees


Jenkinsfile
```diff
@@ -0,0 +1 @@
+buildPlugin(jenkinsVersions: [null, '2.60.1'])
```
jglick: Go for .2!

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new JUnitResultsStepExecution(this, context);
+    }
+
+    @Extension
+    public static class DescriptorImpl extends StepDescriptor {
+        @Override
+        public String getFunctionName() {
+            return "junitResults";
```
jglick: It is going to be very hard to explain to users why they should use this instead of `junit`.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -0,0 +1,61 @@
...
+                tagsAction.addTag(JUnitResultsStep.HAS_TEST_RESULTS_TAG_NAME, "true");
+                node.save();
+            } else {
+                tagsAction = new TagsAction();
+                tagsAction.addTag(JUnitResultsStep.HAS_TEST_RESULTS_TAG_NAME, "true");
+                node.addAction(tagsAction);
+            }
+        }
+
+        return null;
```
jglick: This omits the critical functionality in #9: allowing scripts to interpret results themselves.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
abayer: Yeah, ideally, I'd like to switch `junit` to this, but I'm wary - the parameters are the same, so new builds should be fine, but will currently-running builds at upgrade time have problems?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
abayer: Good point!

abayer:
@jglick Yup, that's the intent. Via `TestResult.getResultsByRunAndNodes`, Blue Ocean will be able to get all the test results for 1..n steps and then display just those test results.

pom.xml
```diff
@@ -13,8 +13,8 @@
     <description>Allows JUnit-format test results to be published.</description>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/JUnit+Plugin</url>
     <properties>
-        <jenkins.version>1.580.1</jenkins.version>
-        <java.level>6</java.level>
+        <jenkins.version>2.7.3</jenkins.version>
```
svanoort: Is there something here that requires 2.7.3?   Might be worth holding it to 1.651 if we can since that has a pretty large install base according to stats.jenkins.io 

I know, I know, it's over the year support maximum by a month or two, but I figure there's no hard in taking an slightly older core dep if we don't need the newer things.

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -241,7 +241,7 @@ public void setData(List<Data> testData) {
     /**
      * Merges an additional test result into this one.
      */
-    void mergeResult(TestResult additionalResult, TaskListener listener) {
+    public void mergeResult(TestResult additionalResult, TaskListener listener) {
```
svanoort: Does it need to be public?

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -142,11 +150,19 @@ protected TestResult parse(String expandedTestResults, AbstractBuild build, Laun
     @Override
     public void perform(Run build, FilePath workspace, Launcher launcher,
             TaskListener listener) throws InterruptedException, IOException {
+        TestResultAction action = parseAndAttach(this, null, build, workspace, launcher, listener);
+
+        if (action != null && action.getResult().getFailCount() > 0)
+            build.setResult(Result.UNSTABLE);
```
svanoort: This doesn't feel right to me -- don't people tend to handle the way test results map to build results in vastly different ways, generally?

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -126,22 +154,23 @@
...
         Element root = result.getRootElement();
 
-        parseSuite(xmlReport,keepLongStdio,r,root);
+        parseSuite(xmlReport,keepLongStdio,r,root, runId, nodeId);
 
         return r;
     }
 
-    private static void parseSuite(File xmlReport, boolean keepLongStdio, List<SuiteResult> r, Element root) throws DocumentException, IOException {
+    private static void parseSuite(File xmlReport, boolean keepLongStdio, List<SuiteResult> r, Element root, String runId,
```
svanoort: Recommend making heavy use of CheckForNull in the runId and nodeId fields/parameters we've added.

svanoort:
This very closely matches the result laid out in [the proposal I laid out on Jenkins-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395?focusedCommentId=256097&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-256097) after some conversations with Abayer, so naturally I like the general approach here. ;)  -- prepare for a batch of comments when review is done though

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -243,6 +277,26 @@ public float getDuration() {
...
+        return runId;
+    }
+
+    /**
+     * The possibly-null {@link FlowNode#id} this suite was generated in.
+     *
+     * @since 1.21
+     */
+    @Exported(visibility=9)
+    public String getNodeId() {
```
svanoort: CheckForNull (here and abvoe)

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -67,6 +71,11 @@
     private transient Map<String,SuiteResult> suitesByName;
 
     /**
+     * {@link #suites} keyed by their run ID and node ID for faster lookup. May be empty.
+     */
+    private transient Map<String,Map<String,List<SuiteResult>>> suitesByRunAndNode;
```
svanoort: [MMMMMMonster-map!](https://www.youtube.com/watch?v=InBZINtS0ec)

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -596,6 +648,30 @@ public SuiteResult getSuite(String name) {
...
+
+    @Nonnull
+    public TestResult getResultByRunAndNodes(@Nonnull String runId, @Nonnull List<String> nodeIds) {
+        TestResult result = new TestResult();
+        if (suitesByRunAndNode.containsKey(runId)) {
+            for (String n : nodeIds) {
+                List<SuiteResult> suites = suitesByRunAndNode.get(runId).get(n);
+                if (suites != null) {
+                    for (SuiteResult s : suites) {
+                        s.setParent(null);
```
svanoort: I'm not entirely clear what the implications of setParent are here -- it feels rather odd to see this in the getter though, since isn't it changing the persistent state?

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -712,6 +798,16 @@ public void freeze(TestResultAction parent) {
             pr.freeze();
     }
 
+    private void addSuiteByRunAndNode(SuiteResult s) {
+        if (suitesByRunAndNode.get(s.getRunId()) == null) {
```
svanoort: This block is rather confusing to follow, because you're doing successive gets and puts, please could you consider restructuring with nesting of conditionals so it's clear where the flow of execution will go?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
svanoort: I, uh, agree with both points.  If we can retrofit the junit step to use the new APIs correctly, I'd prefer that (it's what I'd originally envisioned).  But yeah we'd need to do some serious testing that it won't break things (should be viable using the workflow-cps, etc testing options though?)

pom.xml
abayer: So the main reason for doing so is that there'll be no real benefit for users on <2.7.3 - since the reporting will be in Blue Ocean.

src/main/java/hudson/tasks/junit/TestResultAction.java
abayer: Yup, used over in `xunit`.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
abayer: I waffled on this a lot. This is something that I'd like to differently, but until https://issues.jenkins-ci.org/browse/JENKINS-43995 lands, we don't have a way to pick up that `UNSTABLE` status to cover the whole build later (modulo `try/catch` and the like). So for the moment, I'd say go with principle of least surprise: if you use `junit` now, it sets the build to `UNSTABLE`, so `junitResults` should do the same.

src/main/java/hudson/tasks/junit/SuiteResult.java
abayer: Good point.

src/main/java/hudson/tasks/junit/SuiteResult.java
abayer: *nod*

src/main/java/hudson/tasks/junit/TestResult.java
abayer: Yeaaaaah, I don't love it, but hey, it works.

src/main/java/hudson/tasks/junit/TestResult.java
abayer: Good point. I cargo-culted this from `merge`, since the `TestResult` we return is similar, but it's probably without a point, yeah.

src/main/java/hudson/tasks/junit/TestResult.java
abayer: Yup, it's evolved a bit awkwardly.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
abayer: *nod* I'm not 100% sure how we can test it such that `junit` means different things in a single run (albeit with a restart) but if we can find a way to do that, I'm all in favor.

src/main/java/hudson/tasks/junit/TestResult.java
abayer: Hmm. At least at first glance, I'm honestly not sure how to restructure this. For the moment, I'll add comments and if there's something you or anyone else sees as a better approach to the same end result, I'll happily embrace it.

abayer:
Hmm...anyone know if it's possible to just automatically use the help and `config.jelly` from another class? `JUnitResultsStep` needs pretty much (if not entirely) identical `config.jelly` and `help*.html` to `JUnitResultArchiver`, so I'm wondering if I can avoid copy-paste...

abayer:
btw, this now incorporates stage/branch into the case result names, assuming there's more than one node with tests.

svanoort:
@abayer See https://github.com/jenkinsci/lockable-resources-plugin/blob/e2ef99f0d07daedb75082cb59cef83e24ca33390/src/main/resources/org/jenkins/plugins/lockableresources/LockableResourcesManager/config.jelly#L18-L18 for an example using the config jelly from another class

svanoort:
@abayer Is this the one you were requesting review on?  If so I'm confused because it still appears to be WIP.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -256,10 +262,25 @@ private static String getSkippedMessage(Element testCase) {
...
 
-    public String getDisplayName() {
+    public String getTransformedTestName() {
         return TestNameTransformer.getTransformedName(testName);
     }
 
+    public String getDisplayName() {
+        // Only prepend the enclosing flow node names if there are any and the run this is in has multiple blocks directly containing
+        // test results.
+        if (!getEnclosingFlowNodeNames().isEmpty()) {
```
svanoort: Think long and hard about whether or not to fetch enclosingFlowNode names -- this generally requires iterating through the entirety of the flow, so it is not a trivial operation.  Cheaper to do once and store, sure -- just consider if this information is something you really need.

abayer: I think it is - given that the desired functionality is "tell me what stage/branch/etc this test was reported in", we either need to record the stage/branch/etc information (i.e., the enclosing flow nodes) in the test result, at which point we've got it baked in that one time we write to it and never have to query again, or we need to take the step flow node where the test was recorded and, every time we want to report on tests, determine what stage/branch/etc that step was in.

The two primary "use cases" I've come across for this so far are:
* Getting the "fully qualified" name (including stage/branch) for every test in a run for display purposes, i.e., the Blue Ocean or classic Jenkins test result screens
* Getting the list of tests recorded somewhere within a stage/branch, i.e., to have a more granular version of the existing test result screens scoped to a specific stage/branch

The former is the one I've currently implemented here and in Blue Ocean, fwiw.

Both of those use cases are dramatically easier if enclosing flow nodes are baked into the test result. As demonstrated here with `CaseResult#getDisplayName()`, the only complexity added to the display logic here in `junit` is to check if we need to prepend the stage/branch name(s) at all (i.e., if there are multiple enclosing stages/branches amongst all the test results for this run) - and even that is working entirely from data under the `TestResultAction`. To do this without recording the stage/branch info at runtime, each time we wanted to display the results, we'd have to load the `FlowExecution`, find the `FlowNode` where the test result was recorded, find its enclosing stage/branch, etc - yeah, we could cache that (and would have to in practice, I think, since the alternative would be doing that query path for every single test over and over when we need to check if we should display the stage/branch names in the first place) but it still exposes us to the same costs as doing it at runtime, with the strong possibility of doing it all more than once (i.e., when the cache expires, when the master restarts, etc).

And the second use case...well, we'd have to determine the flow nodes contained within a given stage/branch from the `FlowExecution` and then find all test results recorded in one of those contained flow nodes. And that doesn't seem to be an easy thing to do without some chunk visitor fun, to say the least.

I'd definitely like a more optimized way to say "get me the names, in some predictable order, of any stage or parallel branch containing my current flow node" at runtime.

abayer:
@svanoort Oops, forgot to remove WIP. Sorry.

src/main/java/hudson/tasks/junit/CaseResult.java
svanoort: I agree it's *extremely* useful for the user, I'm just deeply concerned about the performance implications of repeatedly scanning the flowgraph from end to start -- for example, in the common case of parallel test runs, we'd have N branches with M nodes preceding this, and archiving JUnit results would be O(m*n) overall, which would probably be surprising to users. 

Scanning for the nearest enclosing stage can *potentially* exhibit that performance, but in practice that is very where.  Could we extend the cache from https://github.com/jenkinsci/workflow-api-plugin/pull/45 to provide the unclosed blocks?

abayer: I defer to you on that - I'm not nearly well versed enough to say.

abayer:
@jtnord See https://github.com/jenkinsci/xunit-plugin/pull/49 and https://github.com/jenkinsci/pipeline-maven-plugin/pull/82 for examples of other plugins building off this.

The only change I've made here to classic reporting is prepending the stage/branch name(s) to the display name of the `CaseResult`, when appropriate. More work may be needed for trends, though.

src/main/java/hudson/tasks/junit/CaseResult.java
svanoort: I think so.  What we'd have to do is add a mapping from the current head(s) to the enclosing, unterminated block, and then from that block or blocks to the other enclosing blocks.  With each new head, we update that mapping. 

Would look like a mapping of nodeId -> enclosingBlock, in addition to tracking overall unclosed blocks

abayer: If you can put that together, I'll happily hold off on finishing this for it.

jtnord:
@abayer - maybe I was not clear.  hudson.tasks.tests.* is the package used for people wanting to extend the generic test reporting.
hudso.tasks.junit is a junit specific implementation of this.  (they both happen to be in this plugin for historical reasons)

What I am saying is what about those things that build of hudson.tasks.test.*  are they expected to do their own thing?

Nether xunit nor pipeline-maven-plugin are examples of this.

abayer:
@jtnord Aaaah. Lemme think about the `hudson.tasks.test.*` stuff. Good point.

abayer:
@jtnord Ok, I need to do some more experimenting, but that may sadly be a bridge too far - the model is so bare in the `hudson.tasks.test` classes in relation to inheritance etc, and `SuiteResult` is so thoroughly *not* a `hudson.tasks.test.TestResult` that it may not be viable to push this down there. I'll keep experimenting, but it may well be that `cucumber-testresults-plugin` and friends do have to do their own implementation. If that's the case, I'll try to abstract as much of the Pipeline logic out so that it's as simple as possible to implement.

jtnord:
@abayer - how would this API be used in say https://github.com/jenkinsci/parallel-test-executor-plugin
This is a vital piece of the puzzle, can it be validated that the API works for this use case so that it does not have to be re-worked again?

src/main/java/hudson/tasks/junit/CaseResult.java
abayer: @svanoort re the cache you're talking about - do you basically just mean running, say,
```
LinearBlockHoppingScanner scanner = new LinearBlockHoppingScanner();
FlowNode enclosing = scanner.findFirstMatch(this, FlowScanningUtils.MATCH_BLOCK_START);
if (enclosing != null) {
  this.enclosingId = enclosing.getId();
}
```
...at `FlowNode` instantiation time? And then providing something like this on `FlowNode`:
```
    private transient List<FlowNode> enclosingBlocks;

    @Nonnull
    public List<FlowNode> getEnclosingBlocks() {
        if (enclosingBlocks == null) {
            enclosingBlocks = new ArrayList<>();
            if (enclosingId != null) {
                try {
                    FlowNode enclosingNode = exec.getNode(enclosingId);
                    if (enclosingNode != null) {
                        enclosingBlocks.add(enclosingNode);
                        enclosingBlocks.addAll(enclosingNode.getEnclosingBlocks());
                    }
                } catch (IOException e) {
                    LOGGER.log(Level.WARNING, "failed to load parents of " + id, e);
                }
            }
        }
        return enclosingBlocks;
    }
```

abayer: With some tweaks (changing to `scanner.findFirstMatch(getParents(), Collections.singleton(this), FlowScanningUtils.MATCH_BLOCK_START);`, only populating `enclosingId` for `AtomNode` and `BlockStartNode`), this worked. So...if that's the right thing to do, I'll write some more tests and toss up a PR in `workflow-api`.

svanoort: Excellent!  Yes, that's very close to what I was thinking - with just a couple tweaks: 

1. We don't need to do a scan for every new node.  We can just use the mapping for the previous node... after checking if it is itself a block start/end.  If it's a block end, then we get the block start node corresponding to *that* and examine it.  Yay presto, enclosing block in O(1) time for the common case, with a tiny handful of cases where you need to walk further up.
2.  Probably worth having an early exit for FlowStartNode
3. Maybe we want to use String IDs entirely rather than FlowNode objects, when possible?

I doubt any of those actually change the logic materially that much though.  What you've got is more or less exactly what I was thinking!  :man_dancing: 

One note: we may want to ensure the actual implementation itself is never exposed in case (cough) this is supported directly via the FlowNodeStorage implementations in the future.

abayer: So previous node here means `parents`? 

We're kinda stuck using `FlowNode`s for `getEnclosingBlocks()` (rather than string IDs) - well, more accurately, we're going to have to load the `FlowNode` for the immediate enclosing ID to be able to get *its* enclosing block and so on, so if we're going that far, hey, why not just stick that all in a `transient` field and avoid having to do that loading again later.

svanoort: >  here means parents

Yes, more specifically *parent.*   Multiple parents are only possible for the BlockEndNode ending  a set of Parallel branches -- and because that's a block end node, you look at thee BlockStartNode and its parent.

abayer: Well, it would mean `FlowNode.parents`, but it'd always be a max of 1. =) In any case, my current experimentation is only populating the `enclosingId` for `AtomNode` and `BlockStartNode`. I'll get a PR up tomorrow.

svanoort: > so if we're going that far, hey, why not just stick that all in a transient field and avoid having to do that loading again later

Well, so the *other* option is to cache *all* enclosing blocks for a FlowNode in a List<String>.   So you only do a single lookup, then you can load all the FlowNodes. 

Let's say that in the future I anticipate efficient bulk loading of FlowNodes will become A Thing and leave it at that for now. ;)

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new JUnitResultsStepExecution(this, context);
+    }
+
+    @Extension
+    public static class DescriptorImpl extends StepDescriptor {
+        @Override
+        public String getFunctionName() {
+            return "junit";
```
jglick: You need to pick something different, as this will clash with the symbol aliased to `step([$class: 'JUnitResultsArchiver', …])`.

abayer: Check elsewhere in the PR - I switched that symbol.

abayer: Check elsewhere in the PR - I switched that symbol.

abayer: Check elsewhere in the PR - I switched that symbol.

pom.xml
```diff
@@ -64,5 +88,52 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-cps</artifactId>
+            <version>${workflow-cps.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-job</artifactId>
+            <version>2.11.1</version>
```
jtnord: pipeline deps should be `optional`... whilst I have no reason not to install pipeline, just because I use junit I should not require it.

abayer: I thought about that, but decided we're far enough along in the Pipeline era now that it's eminently reasonable to assume that Pipeline be installed everywhere. =)

pom.xml
nfalco79: Not really. In my company for example we do not use Blue Ocean because we use scripted pipeline and we have not plan conversion.
Our project has flow with two parallel stages that run integration tests on two different application server. Our needs is differentiate tests to understand from which step comes to understand on which WAS the test is failed.
Now we do know unless to archive surefire reports (xml) and check manually.

pom.xml
nfalco79: I agree with @jtnord , you should not mark as not optional dependency only because you assume pipeline is installed everywhere because with this trend jenkins became a big monolite where installing one plugin downloads all.

jtnord: @abayer NO NO.  ping me offline if you want to be educated why pipeline will never be installed on ALL jenkins instances, and why doing this in junit is pure evil.

abayer:
Aaaaahhh, good point re split from core. Will change to optional.

abayer:
@jtnord @jglick @svanoort Ok, this is now caught up to use `workflow-api` 2.22 with the new API for enclosing block info, and Pipeline dependencies other than `workflow-step-api` are optional.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -0,0 +1,83 @@
...
+     * Get the stage and parallel branch start node IDs (not the body nodes) for this node, innermost first.
+     * @param node A flownode.
+     * @return A nonnull, possibly empty list of stage/parallel branch start node IDs, innermost first.
+     */
+    @Nonnull
+    public static List<String> getEnclosingStagesAndParallels(FlowNode node) {
+        List<String> enclosingBlocks = new ArrayList<>();
+        for (FlowNode enclosing : node.getEnclosingBlocks()) {
+            if (enclosing != null && enclosing.getAction(LabelAction.class) != null) {
+                if ((enclosing instanceof StepStartNode && ((StepStartNode) enclosing).getDescriptor() instanceof StageStep.DescriptorImpl) ||
```
jtnord: will this not blow up if `pipeline-stage-step` is not installed (it is optional)...

abayer: Yes. Either we make the pipeline dependencies optional and accept that things blow up (`FlowNode` from `workflow-api` too!) or we make them non optional.

nfalco79: Move the specific pipeline step executor to a separate plugin like
* docker-build-step-plugin
* workflow-step-api-plugin
* pipeline-maven-plugin
* XYZ-build-step-api-plugin
and so on

abayer: ...why?

nfalco79: for the same reason those plugin are created. Move pipeline dependency far from base plugin. For the same reason do not force dependency from Jenkins 2.7.3 because
> So the main reason for doing so is that there'll be no real benefit for users on <2.7.3 - since the reporting will be in Blue Ocean.
than move this step to one of Blue Ocean plugin.

abayer: Those plugins don’t duplicate existing behavior, though. This is the right place for a junit step.

nfalco79: A similar concept was expressed by jglick [here](https://github.com/jenkinsci/junit-plugin/pull/9#issuecomment-68792525)

abayer: That was 2.5 years ago - a lot has changed since then, most notably the end of bundling.

jtnord: @abayer so this should not blow up.  If you can not use this step without the `pipeline-stage-step` then the step should not be active (ie you can not use it), or you can use it but you get no "per stage" results.

You should be able to do this, there are many ways to do it.
one would be to check if the plugin is installed and only then call some code in a different class.
I'm not sure if Steps can be marked as Optional using the varient plugin or the like to say this is only active if this plugin is installed, but I guess it should be possible, if not I think we should be able to do this.

pom.xml
```diff
@@ -44,7 +47,30 @@
...
+            <version>2.22</version>
+            <optional>true</optional>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                    <artifactId>scm-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
```
jtnord: just wondering why you need script-security - you don;t call any groovy...

pom.xml
```diff
@@ -64,5 +90,55 @@
             <version>1.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-cps</artifactId>
+            <version>${workflow-cps.version}</version>
+            <optional>true</optional>
```
jtnord: same here - you do not seem to need to depend on this, you just want an optional on workflow job do you not?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
```diff
@@ -0,0 +1,158 @@
...
+    @DataBoundSetter public final void setAllowEmptyResults(boolean allowEmptyResults) {
+        this.allowEmptyResults = allowEmptyResults;
+    }
+
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new JUnitResultsStepExecution(this, context);
+    }
+
+    @Extension
```
jtnord: so you currently require "pipeline-stage-step" as optional - so this should be marked as optional depending on the installation of that plugin - or you need to make the code not blow up.

Check the [Varient plugin](https://wiki.jenkins.io/display/JENKINS/Variant+Plugin)

abayer: nifty!

pom.xml
abayer: `hudson.tasks.junit.TestResultSummary` uses the `@Whitelisted` annotation.

pom.xml
abayer: `hudson.tasks.junit.pipeline.JUnitResultsStepExecution` uses `org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode`

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
abayer: Going with the `variant` approach.

pom.xml
svanoort: Stats have shifted enough since July that I think we're good to bump to 2.7.3 without issue.  There's still some legacy users but I'm less concerned about them getting new features added to this plugin after that.

pom.xml
```diff
@@ -44,7 +47,34 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.2</version>
+            <version>1.7</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <version>2.12</version>
```
svanoort: Was pipeline going to be an optional dependency or not?  

I'm confused about what all has happened here since I built the APIs to let you fetch enclosing blocks for this.

svanoort: Maybe this one just missed an optional?

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -235,14 +299,36 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio) throws
...
+                    if (node.getAction(LabelAction.class) != null) {
+                        ThreadNameAction threadNameAction = node.getAction(ThreadNameAction.class);
+                        if (threadNameAction != null) {
+                            return threadNameAction.getThreadName();
+                        }
+                    }
+                    return node.getDisplayName();
+                }
+            } catch (Exception e) {
+                // TODO: Something in that odd case where we can get an NPE
```
svanoort: Wait, NPE where?  It looks like this is null protected?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
svanoort: I don't understand the variant approach but if it deals with this I'm all for it.

svanoort: Actually @abayer  you *could* just have it look for the descriptor *ID* matching, which means you don't need a hard dependency on stage-step (string equals vs instanceof).

svanoort:
@abayer  I'm unclear if this one is ready for review again, but it looks pretty reasonable to me modulo some funkiness around dependency handling.

src/main/java/hudson/tasks/junit/SuiteResult.java
abayer: `execution.getNode` can theoretically throw an NPE. =)

abayer:
@svanoort Yeah, I think this is done, pending @jtnord's re-signoff. Will merge and cut a beta at that point.

pom.xml
```diff
@@ -44,7 +47,34 @@
...
+                <exclusion>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                    <artifactId>scm-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <version>1.30</version>
```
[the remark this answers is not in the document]
abayer: Not historically - `hudson.tasks.junit.TestResultSummary`, which is added here for having something we can return from a `junit` step call, needs `@Whitelisted`. The class isn't an `Extension`, so my gut reaction is that we shouldn't do an optional dependency, but if I'm wrong I'm happy to change it.

pom.xml
batmat: > Not really. In my company for example we do not use Blue Ocean because we use scripted pipeline and we have not plan conversion.

@nfalco79 Sorry, but I don't see the relationship, FWIW. You can (and we definitely do) use Blue Ocean and it will render *any* Pipeline, be it Scripted or Declarative.
Sorry to digress, but I thought I wouldn't leave that statement without an answer for possibly passing-by people.

nfalco79: @batmat that i remember blue ocean can render scripted pipeline only if non "standard" steps are written into a sort of step wrapper. And if i remember rigth also statements like if/def variable or function/for cycles and other groovy commands cause error in the parser when I had open the page of the our pipeline. So I had remove blue ocean and 15 related plugin after 3 hours.

batmat: Well, not denying you could have hit some bug. If you're able to report it, that would be great. But yes, we use Blue Ocean in many many places, definitely too using Scripted pipelines, so this does work normally. Not being the case looks like this would be great to be reported somewhere.

PR #77, opened 2017-07-28
Align headers where fields are aligned

olivergondza:
We are aligning values to the right but column headers are centered. It can get tricky to navigate in bigger tables.

Before:
junit-orig

After:
junit-new


recena:
👍 Although it would be nice to move these inline CSS style to a CSS file into the plugin itself.

PR #79, opened 2017-09-02
Move BlueJUnitTestResult to JUnit plugin

i386:
@jglick as discussed this afternoon. Tests currently fail because this implementation is already in Blue Ocean 1.2 (you end up with duplicate tests. 

We may have to do some sort of dance to make sure that we release 1.3.x without its JUnit dependency and then publish. PR for blueocean-plugin coming later.


abayer:
I'm not keen on this - creates a circular dependency...

timja:
Closing due to inactivity and unresolved issues for over 2 years

PR #83, opened 2017-10-19
[JENKINS-27395] Removed per-run node tracking

abayer:
Yet more for [JENKINS-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395)

On further inspection, I came to the conclusion that since
multiple-runs-per-TestResult was only possible for
`AggregatedTestResultAction`, which doesn't support Pipeline jobs in
the first place, it made more sense to simplify things and get rid of
the run-and-nodes tracking of suites in favor of just nodes.

This also necessitated gathreing the enclosing block names earlier,
but that's probably a good change anyway.

cc @reviewbybees

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -41,7 +41,7 @@ import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Publisher;
 import hudson.tasks.Recorder;
 import hudson.tasks.junit.TestResultAction.Data;
-import hudson.tasks.test.PipelineArgs;
+import hudson.tasks.test.PipelineTestDetails;
 import hudson.util.DescribableList;
 import hudson.util.FormValidation;
 import org.apache.tools.ant.DirectoryScanner;
@@ -124,16 +124,16 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
     private TestResult parse(String expandedTestResults, Run<?,?> run, @Nonnull FilePath workspace, Launcher launcher, TaskListener listener)
             throws IOException, InterruptedException
     {
-        return parse(this, new PipelineArgs(), expandedTestResults, run, workspace, launcher, listener);
+        return parse(this, null, expandedTestResults, run, workspace, launcher, listener);
 
     }
 
-    private static TestResult parse(@Nonnull JUnitTask task, @Nonnull PipelineArgs pipelineArgs,
+    private static TestResult parse(@Nonnull JUnitTask task, PipelineTestDetails pipelineTestDetails,
                                     String expandedTestResults, Run<?,?> run, @Nonnull FilePath workspace,
                                     Launcher launcher, TaskListener listener)
             throws IOException, InterruptedException {
         return new JUnitParser(task.isKeepLongStdio(), task.isAllowEmptyResults())
-                .parseResult(expandedTestResults, run, pipelineArgs, workspace, launcher, listener);
+                .parseResult(expandedTestResults, run, pipelineTestDetails, workspace, launcher, listener);
     }
 
     @Deprecated
@@ -150,20 +150,20 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
     @Override
     public void perform(Run build, FilePath workspace, Launcher launcher,
             TaskListener listener) throws InterruptedException, IOException {
-        TestResultAction action = parseAndAttach(this, new PipelineArgs(), build, workspace, launcher, listener);
+        TestResultAction action = parseAndAttach(this, null, build, workspace, launcher, listener);
 
         if (action != null && action.getResult().getFailCount() > 0)
             build.setResult(Result.UNSTABLE);
     }
 
-    public static TestResultAction parseAndAttach(@Nonnull JUnitTask task, @Nonnull PipelineArgs pipelineArgs,
+    public static TestResultAction parseAndAttach(@Nonnull JUnitTask task, PipelineTestDetails pipelineTestDetails,
                                                   Run build, FilePath workspace, Launcher launcher, TaskListener listener)
             throws InterruptedException, IOException {
         listener.getLogger().println(Messages.JUnitResultArchiver_Recording());
 
         final String testResults = build.getEnvironment(listener).expand(task.getTestResults());
 
-        TestResult result = parse(task, pipelineArgs, testResults, build, workspace, launcher, listener);
+        TestResult result = parse(task, pipelineTestDetails, testResults, build, workspace, launcher, listener);
 
         synchronized (build) {
             // TODO can the build argument be omitted now, or is it used prior to the call to addAction?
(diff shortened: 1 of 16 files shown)


abayer:
ping @jtnord @jglick @svanoort - would like to get this reviewed and merged so that the downstream PRs can get reviewed and we can cut another beta. =)

abayer:
@jtnord Yup, there hasn't been a non-beta release of 1.22 yet, so we should be good to go.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -59,22 +61,45 @@ protected TestResultSummary run() throws Exception {
...
-    public static List<String> getEnclosingStagesAndParallels(FlowNode node) {
-        List<String> enclosingBlocks = new ArrayList<>();
+    public static List<FlowNode> getEnclosingStagesAndParallels(FlowNode node) {
+        List<FlowNode> enclosingBlocks = new ArrayList<>();
         for (FlowNode enclosing : node.getEnclosingBlocks()) {
             if (enclosing != null && enclosing.getAction(LabelAction.class) != null) {
                 if ((enclosing instanceof StepStartNode && ((StepStartNode) enclosing).getDescriptor() instanceof StageStep.DescriptorImpl) ||
                         (enclosing.getAction(ThreadNameAction.class) != null)) {
-                    enclosingBlocks.add(enclosing.getId());
+                    enclosingBlocks.add(enclosing);
```
svanoort: 🐛  Why aren't we getting the ID?

abayer: Because I decided to change this to return a list of `FlowNode`s instead - pretty much every time I was calling it, I wanted the `FlowNode`s rather than just the IDs, so hey.

svanoort: D'oh, misread that.  Okay, approving!

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -150,21 +150,21 @@ protected TestResult parse(String expandedTestResults, AbstractBuild build, Laun
...
+        TestResultAction action = parseAndAttach(this, null, null, null, build, workspace, launcher, listener);
 
         if (action != null && action.getResult().getFailCount() > 0)
             build.setResult(Result.UNSTABLE);
     }
 
     public static TestResultAction parseAndAttach(@Nonnull JUnitTask task, @CheckForNull String nodeId,
-                                                  List<String> enclosingBlocks, Run build, FilePath workspace,
-                                                  Launcher launcher, TaskListener listener)
+                                                  List<String> enclosingBlocks, List<String> enclosingBlockNames,
```
jglick: This does not seem like a sustainable API. Introduce a new (mutable-until-“frozen”? builder?) struct type for the arguments, so that we can compatibly add information to the parser without needing to manipulate massive argument lists like this.

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -347,20 +318,9 @@ public float getDuration() {
...
-    @Exported(visibility=9)
-    public String getRunId() {
-        return runId;
-    }
-
-    /**
      * The possibly-null {@link FlowNode#id} this suite was generated in.
      *
-     * @since 1.21
+     * @since 1.22
```
jglick: BTW I really do not recommend merging API stuff into `master` until the whole set of related changes has been fully implemented and approved.

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -135,15 +135,15 @@ public TestResult(long buildTime, DirectoryScanner results, boolean keepLongStdi
...
      * @param nodeId an optional {@link org.jenkinsci.plugins.workflow.graph.FlowNode#id} for mapping between suite results and step executions.
      * @param enclosingBlocks Optional, possibly null list of enclosing {@link FlowNode#getId()}
-     * @since 1.21
+     * @param enclosingBlockNames Optional, possibly null list of enclosing block names
+     * @since 1.22
      */
-    public TestResult(long buildTime, DirectoryScanner results, boolean keepLongStdio, @CheckForNull String runId,
-                      @CheckForNull String nodeId, @CheckForNull List<String> enclosingBlocks) throws IOException {
+    public TestResult(long buildTime, DirectoryScanner results, boolean keepLongStdio, @CheckForNull String nodeId,
+                      @CheckForNull List<String> enclosingBlocks, @CheckForNull List<String> enclosingBlockNames) throws IOException {
```
jglick: Same kind of issue here—feels like there needs to be some API type defining what information the parser needs, so we do not have to forever be doing a dance of deprecating constructors and methods and delegating to similar-looking overloads.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -59,22 +61,45 @@ protected TestResultSummary run() throws Exception {
...
+        }
+        return ids;
+    }
+
+    @Nonnull
+    public static List<String> getEnclosingBlockNames(@Nonnull List<FlowNode> nodes) {
+        List<String> names = new ArrayList<>();
+        for (FlowNode n : nodes) {
+            ThreadNameAction threadNameAction = n.getAction(ThreadNameAction.class);
+            if (n.getAction(LabelAction.class) != null && threadNameAction != null) {
```
jglick: Why do you need to check for `LabelAction` here?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -59,22 +61,45 @@ protected TestResultSummary run() throws Exception {
...
+                names.add(threadNameAction.getThreadName());
+            } else {
+                names.add(n.getDisplayName());
```
jglick: Not `LabelAction`? Weird. Maybe I do not understand how you are identifying results but this looks suspicious.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
abayer: Good question. Can't remember if there was a reason...

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
abayer: Could be changed to `LabelAction`, but we've already filtered down the enclosing blocks to either `stage` or `parallel`, and the `else` case here means it's gotta be a `stage`, so `getDisplayName()` works.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
abayer: *nod*

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
abayer: Anyway, refactoring this a bit...

abayer:
@jglick @svanoort Any chance of a last re-review?

svanoort:
@abayer Why not just accept null pipelineArgs rather than always creating empty ones?

abayer:
@svanoort Good question. I tend to prefer avoiding null checks if I can help it, but that would simplify things. Off I go!

PR #87, opened 2017-11-10
Make script-security dependency optional

abayer:
There's no scenario I can come up with where the `@Whitelisted` annotation on `TestResultSummary` would be relevant if you don't have `script-security` installed, so let's just make `script-security` optional like all the Pipeline plugins.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
             <version>1.30</version>
+            <optional>true</optional>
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Some explanation in the PR would be helpful

recena:
LGTM. @oleg-nenashev I think this was added previously in a more restrictive scope when no needed.

abayer:
@recena @jtnord Pinging again for review so I can merge and release. =)

jglick:
No, stop using `optional` dependencies!

PR #88, opened 2017-11-15
Simplified dependencies

jglick:
Optional dependencies are evil and variants are eviller. KISS

Solves ATH regressions in `WorkflowPluginTest.{linearFlow,parallelTests}`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.36</version>
+        <version>3.0</version>
         <relativePath />
     </parent>
     <artifactId>junit</artifactId>
@@ -55,15 +55,9 @@
             <version>2.12</version>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>variant</artifactId>
-            <version>1.1</version>
-        </dependency>
-        <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
             <version>2.22</version>
-            <optional>true</optional>
             <exclusions>
                 <exclusion>
                     <groupId>org.jenkins-ci.plugins</groupId>
@@ -99,13 +93,13 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <version>${workflow-cps.version}</version>
-            <optional>true</optional>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
             <version>2.11.1</version>
-            <optional>true</optional>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -137,7 +131,7 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-stage-step</artifactId>
             <version>2.2</version>
-            <optional>true</optional>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,6 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
             <version>1.30</version>
-            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
--- a/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
+++ b/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
@@ -1,6 +1,7 @@
 package hudson.tasks.junit.pipeline;
 
 import com.google.common.collect.ImmutableSet;
+import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.Descriptor;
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -99,13 +92,13 @@
...
             <version>${workflow-cps.version}</version>
-            <optional>true</optional>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
             <version>2.11.1</version>
-            <optional>true</optional>
+            <scope>test</scope>
```
jglick: were not using it as of #83 anyway

pom.xml
```diff
@@ -75,7 +69,6 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
             <version>1.30</version>
-            <optional>true</optional>
```
jglick: reverting #87

pom.xml
```diff
@@ -99,13 +92,13 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <version>${workflow-cps.version}</version>
-            <optional>true</optional>
+            <scope>test</scope>
```
jglick: Was never necessary in #76.

pom.xml
```diff
@@ -137,7 +130,7 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-stage-step</artifactId>
             <version>2.2</version>
-            <optional>true</optional>
+            <scope>test</scope>
```
jglick: see below

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
```diff
@@ -124,7 +124,7 @@ public StepExecution start(StepContext context) throws Exception {
         return new JUnitResultsStepExecution(this, context);
     }
 
-    @OptionalExtension(requirePlugins = {"pipeline-stage-step","workflow-cps","workflow-job"})
+    @Extension
```
jglick: Must always be enabled. Otherwise there is no `junit` step on installations without `pipeline-stage-step`! Ouch.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -70,7 +69,7 @@ protected TestResultSummary run() throws Exception {
         List<FlowNode> enclosingBlocks = new ArrayList<>();
         for (FlowNode enclosing : node.getEnclosingBlocks()) {
             if (enclosing != null && enclosing.getAction(LabelAction.class) != null) {
-                if ((enclosing instanceof StepStartNode && ((StepStartNode) enclosing).getDescriptor() instanceof StageStep.DescriptorImpl) ||
+                if ((enclosing instanceof StepNode && ((StepNode) enclosing).getDescriptor().getFunctionName().equals("stage")) ||
```
jglick: Removing two unnecessary deps here. Cf. suggestions in #76 by both @svanoort and @jtnord, and my requests in https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r145182474 and https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r149539509.

(Also https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r145182600: I am not sure why we even care whether this is a `stage` or something else with `LabelAction`? And why the separate check for `ThreadNameAction` when `parallel` block nodes will _also_ have `LabelAction`? Cf. https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r149539657. Anyway, here I am not changing behavior, just removing unnecessary code.)

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -70,7 +69,7 @@ protected TestResultSummary run() throws Exception {
         List<FlowNode> enclosingBlocks = new ArrayList<>();
         for (FlowNode enclosing : node.getEnclosingBlocks()) {
             if (enclosing != null && enclosing.getAction(LabelAction.class) != null) {
-                if ((enclosing instanceof StepStartNode && ((StepStartNode) enclosing).getDescriptor() instanceof StageStep.DescriptorImpl) ||
+                if (isStageNode(enclosing) ||
```
jglick: Removing two unnecessary deps here. Cf. suggestions in #76 by both @svanoort and @jtnord, and my requests in https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r145182474 and https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r149539509.

(Also https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r145182600: I am not sure why we even care whether this is a `stage` or something else with `LabelAction`? And why the separate check for `ThreadNameAction` when `parallel` block nodes will _also_ have `LabelAction`? Cf. https://github.com/jenkinsci/parallel-test-executor-plugin/pull/31#discussion_r149539657. Anyway, here I am not changing behavior, just removing unnecessary code.)

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
jtnord: should have just removed the pipeline-stage step then.

pom.xml
```diff
@@ -54,16 +54,10 @@
...
-        <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>variant</artifactId>
-            <version>1.1</version>
-        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
             <version>2.22</version>
-            <optional>true</optional>
```
jtnord: @jglick why did you do this without removiong the exclusion on scm-api.

This breaks any consumer plugin of junit plugin and forces them to declare (an uneeded) dependency on scm-api so that hpi:run and the basic InjectedTest start the plugin works.

PR #89, opened 2017-11-15
[FIXED JENKINS-48020] Show stage in failed tests, fix warnings

abayer:
[JENKINS-48020](https://issues.jenkins-ci.org/browse/JENKINS-48020)

First, this changes `CaseResult#getFullDisplayName()` to include
stage/branch names like `CaseResult#getDisplayName()`, so that the
full information will show up in lists of failed tests, for example.

Second, stop using `CaseResult#getDisplayName()` for indexing
`CaseResult`s in `SuiteResult`. That method will spam warnings if
called too early in the run now due to `CaseResult#getRun()` being
called before the `SuiteResult`'s parent is set.

Third, fix `JUnitResultsStepTest#testTrends` to actually test what
it's supposed to, to fail if it doesn't find the expected display
names, to test full display names as well, and to use the same
filename for the test results each time due to one of the stages' test
file not containing an explicit suite name, leading to its suite's
name being set to the filename, breaking `SuiteResult#getPreviousResult()`.

cc @reviewbybees

### DIFF
--- a/src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
+++ b/src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
@@ -42,6 +42,7 @@ import java.util.List;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
 
 public class JUnitResultsStepTest {
     @Rule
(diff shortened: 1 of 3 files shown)


src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -245,26 +246,36 @@ public void testTrends() throws Exception {
         WorkflowJob j = rule.jenkins.createProject(WorkflowJob.class, "testTrends");
         j.setDefinition(new CpsFlowDefinition("node {\n" +
                 "  stage('first') {\n" +
-                "    def first = junit(testResults: \"junit-report-testTrends-first-${env.BUILD_NUMBER}.xml\")\n" +
+                "    def first = junit(testResults: \"junit-report-testTrends-first.xml\")\n" +
```
jglick: easier to use `'` than `"` now

jglick: also this could be simplified all the way down to

```groovy
junit 'junit-report-testTrends-first.xml'
```

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -245,26 +246,36 @@ public void testTrends() throws Exception {
...
-            firstFile.copyFrom(JUnitResultsStepTest.class.getResource("junit-report-testTrends-first-" + i + ".xml"));
-            FilePath secondFile = ws.child("junit-report-testTrends-second-" + i + ".xml");
-            secondFile.copyFrom(JUnitResultsStepTest.class.getResource("junit-report-testTrends-second-" + i + ".xml"));
-        }
+        FilePath firstFile = ws.child("junit-report-testTrends-first.xml");
+        FilePath secondFile = ws.child("junit-report-testTrends-second.xml");
+
+        // Populate first run's tests.
+        firstFile.copyFrom(JUnitResultsStepTest.class.getResource("junit-report-testTrends-first-1.xml"));
+        secondFile.copyFrom(JUnitResultsStepTest.class.getResource("junit-report-testTrends-second-1.xml"));
```
jglick: Maybe it would be handy for there to be a `Step` defined in some `*-tests.jar` that copied a classpath resource to the workspace, so that tests like this would not need to call `getWorkspaceFor` and could use something like

```groovy
stage('first') {
  writeTestResourceToWorkspace resource: '/some/pkg/junit-report-testTrends-first-1.xml', file: 'junit-report-testTrends-first.xml'
  junit 'junit-report-testTrends-first.xml'
}

PR #92, opened 2017-11-23
[FIXED JENKINS-48196] Remove redundant branch names from case displays

abayer:
[JENKINS-48196](https://issues.jenkins-ci.org/browse/JENKINS-48196)

If we've got a suite with a stage inside a parallel branch where the
stage and parallel branch have the same name, we should only include
that name once in the suite/case/etc display name - it's redundant to
display it twice.

There is a potential gotcha here if you've got multiple stages inside
a branch and one of those stages has the same name as the branch, but
I think that's an edge case we can live with for now.

cc @reviewbybees

### DIFF
--- a/src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
+++ b/src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
@@ -28,6 +28,7 @@ import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.TestExtension;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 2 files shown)


michaelneale:
nice - look ok @rsandell ? (sam is enjoying turkey)

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -327,17 +365,22 @@ public boolean apply(@Nullable FlowNode input) {
...
 
-    public static void assertBranchResults(WorkflowRun run, int suiteCount, int testCount, int failCount, String branchName, String stageName) {
+    public static void assertBranchResults(WorkflowRun run, int suiteCount, int testCount, int failCount, String branchName, String stageName,
+                                           String innerStageName) {
         FlowExecution execution = run.getExecution();
         DepthFirstScanner scanner = new DepthFirstScanner();
         FlowNode aBranch = scanner.findFirstMatch(execution, branchForName(branchName));
         assertNotNull(aBranch);
         TestResult branchResult = assertBlockResults(run, suiteCount, testCount, failCount, aBranch);
+        String namePrefix = stageName + " / " + branchName;
```
rsandell: _nit_ You could have just made `String branchName, String stageName, String innerStageName` into a `String... names`
and then just `Functions.joinPath(names)`

abayer: Except the order there would need to be `stageName`, `branchName`, `innerStageName`, and I need `branchName` on its own above too. I'll come through and refactor my test utilities here eventually, though.

rsandell: ok, hence me approving beforehand ;)

PR #94, opened 2017-12-04
[FIXED JENKINS-48372] Allow ignoring enclosing nodes for results.

abayer:
[JENKINS-48372](https://issues.jenkins-ci.org/browse/JENKINS-48372)

If `ignoreEnclosingDepth` is specified, that many enclosing nodes,
starting from the innermost, are ignored when recording test
results. This allows, for example, not recording parallel branches for
use with parallel-test-executor.

cc @reviewbybees

### DIFF
--- a/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-ignoreEnclosingDepth.html
+++ b/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-ignoreEnclosingDepth.html
@@ -1,0 +1,36 @@
+<div>
+    Optionally, specify the number of enclosing stages or branches, starting from the innermost stage or branch, to
+    ignore. Defaults to 0, meaning all are included.
+
+    Example:
+    <code>
+        stage('outer stage') {
+          stage('inner stage') {
+            parallel(branchA: {
+                       node {
+                         sh './run_first_set_of_tests.sh`
+                         junit testResults: '**/target/surefire-reports/TEST-*.xml'
+                       }
+                     },
+                     branchB: {
+                       node {
+                         sh './run_second_set_of_tests.sh`
+                         junit testResults: '**/target/surefire-reports/TEST-*.xml', ignoreEnclosingDepth: 1
+                       }
+                     },
+                     branchC: {
+                       node {
+                         sh './run_third_set_of_tests.sh`
+                         junit testResults: '**/target/surefire-reports/TEST-*.xml', ignoreEnclosingDepth: 2
+                       }
+                     })
+          }
+        }
+    </code>
+    Tests in <code>branchA</code>, with no <code>ignoreEnclosingDepth</code> specified, will be listed like
+    <code>outer stage / inner stage / branchA / testName</code>.
+    Tests in <code>branchB</code>, with <code>ignoreEnclosingDepth</code> specified as 1, will be listed like
+    <code>outer stage / inner stage / testName</code>, ignoring their innermost enclosing stage or branch.
+    Tests in <code>branchC</code>, with <code>ignoreEnclosingDepth</code> specified as 2, will be listed like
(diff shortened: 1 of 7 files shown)


src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-ignoreEnclosingDepth.html
```diff
@@ -0,0 +1,36 @@
...
+                     branchB: {
+                       node {
+                         sh './run_second_set_of_tests.sh`
+                         junit testResults: '**/target/surefire-reports/TEST-*.xml', ignoreEnclosingDepth: 1
+                       }
+                     },
+                     branchC: {
+                       node {
+                         sh './run_third_set_of_tests.sh`
+                         junit testResults: '**/target/surefire-reports/TEST-*.xml', ignoreEnclosingDepth: 2
```
jglick: This seems rather baroque.

Is it too late to turn off the automatic association of results to enclosing blocks, which in my experience just winds up producing weirdly named test sets, and make the test set name be a simple parameter to the `junit` step as in the original proposal? Then we would not need any tricks like this.

abayer: It is a bit baroque. I do think the automatic logic is valuable, but it may make sense to only pay attention to `parallel`s if a flag is given (or not at all). @i386, @michaelneale - thoughts?

svanoort:
Yeah, I've been deferring an up/down vote while thinking about this one.  It feels like we need to take a step back from the implementation a little and ponder what the desired user experience is and start from that, since this is primarily a UX-driven feature.

abayer:
@michaelneale What I'm spitballing is either never including `parallel` in the list of enclosing blocks for a `junit` call, so that the display name for tests in either branch of `stage('foo') { parallel(a: { ... }, b: { ... } }` would be `foo / testName`, not `foo / a / testName` or `foo / b / testName` (which is particularly annoying with `parallel-test-executor`'s N number of branches that have no semantic meaning, purely functional, with tests that could end up in differently named parallel branches from one run to the next for no user-controlled reason). Optionally, we could also add a flag to `junit` to include "bare" parallels in the display name anyway.

Alternatively, @svanoort is suggesting just a simple "ignore innermost enclosing `parallel` block" boolean flag, rather than the "ignore the first X enclosing blocks, either `parallel` or `stage`" that I've got in the PR currently.

I'm not gonna make an executive decision on this one - would definitely appreciate thoughts.

svanoort:
@michaelneale Specifically, the concrete user story here could be more briefly described as "when using parallel test executor, I want all the test results to behave like I ran one test suite but faster", i.e. why a collapseParallel to ignore the enclosing parallel is helpful.  

Docs would want to indicate explicitly what it's used for and that you normally want it off.

michaelneale:
Ah ok. I like having the option but I don’t know what default should be. @i386 have any opinion?

nimbuscrenel:
Any traction on this? This issue is currently wreaking havoc on how we link to parallel execution test results, and I hate to have to roll back to 1.21.

gcarnell:
Not sure what the status of this is, but I would cast my vote for the default behavior to ignore the enclosing parallel.
I basically want previous functionality back, and @svanoort perfectly summed up my "user story".

jglick:
@abayer what is the status of this?

src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-ignoreEnclosingDepth.html
jglick: > as in the original proposal

At this point I am not sure if I was referring to https://github.com/jenkinsci/junit-plugin/pull/64#issuecomment-281700612 or something else.

timja:
Closing due to inactivity.

PR #95, opened 2017-12-04
Ensure html help for fields for junit step exist

abayer:
Gotta copy them over from `JUnitResultArchiver` to `JUnitResultsStep`,
sadly, or
https://jenkins.io/doc/pipeline/steps/junit/#junit-archive-junit-formatted-test-results
doesn't work right.

cc @reviewbybees


src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-allowEmptyResults.html
```diff
@@ -0,0 +1 @@
+help-allowEmptyResults.html
```
jglick: Huh??

abayer: Welp, symlinking didn't work. Whoops.

PR #96, opened 2017-12-07
[JENKINS-43995] Set FlowNode status when unstable.

abayer:
[JENKINS-43995](https://issues.jenkins-ci.org/browse/JENKINS-43995)

Downstream of https://github.com/jenkinsci/workflow-cps-plugin/pull/192 and https://github.com/jenkinsci/workflow-api-plugin/pull/63

Also tweaked tests to always check against fail count (and verify block status).


jglick:
@abayer what is the status of this?

timja:
> @abayer what is the status of this?

Going to assume abandoned.

PR #98, opened 2017-12-21
Adding a new option to make the build UNSTABLE or not

nguoianphu:
Some people are looking for an option to make the JUnit plug doesn't make the build UNSTABLE even there're failed tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
         <relativePath />
     </parent>
     <artifactId>junit</artifactId>
-    <version>1.24-SNAPSHOT</version>
+    <version>1.25</version>
     <packaging>hpi</packaging>
     <name>JUnit Plugin</name>
     <description>Allows JUnit-format test results to be published.</description>
--- a/src/main/java/hudson/tasks/junit/JUnitParser.java
+++ b/src/main/java/hudson/tasks/junit/JUnitParser.java
@@ -51,11 +51,12 @@ public class JUnitParser extends TestResultParser {
 
     private final boolean keepLongStdio;
     private final boolean allowEmptyResults;
+    private final boolean makeUnstable;
 
     /** TODO TestResultParser.all does not seem to ever be called so why must this be an Extension? */
     @Deprecated
     public JUnitParser() {
-        this(false, false);
+        this(false, false, true);
     }
 
     /**
@@ -66,16 +67,20 @@ public class JUnitParser extends TestResultParser {
     public JUnitParser(boolean keepLongStdio) {
         this.keepLongStdio = keepLongStdio;
         this.allowEmptyResults = false;
+        this.makeUnstable = true;
+    
     }
 
     /**
      * @param keepLongStdio if true, retain a suite's complete stdout/stderr even if this is huge and the suite passed
      * @param allowEmptyResults if true, empty results are allowed
+     * @param makeUnstable if true, build will be Unstable if there are any failed test cases
      * @since 1.10
      */
-    public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults) {
+    public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults, boolean makeUnstable) {
         this.keepLongStdio = keepLongStdio;
         this.allowEmptyResults = allowEmptyResults;
+        this.makeUnstable = makeUnstable;
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -7,7 +7,7 @@
         <relativePath />
     </parent>
     <artifactId>junit</artifactId>
-    <version>1.24-SNAPSHOT</version>
+    <version>1.25</version>
```
oleg-nenashev: it needs to be reverted, the version should be SNAPSHOT in order to let the Maven Release plugin to operate correctly

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -66,16 +67,20 @@ public JUnitParser() {
...
     }
 
     /**
      * @param keepLongStdio if true, retain a suite's complete stdout/stderr even if this is huge and the suite passed
      * @param allowEmptyResults if true, empty results are allowed
+     * @param makeUnstable if true, build will be Unstable if there are any failed test cases
      * @since 1.10
      */
-    public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults) {
+    public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults, boolean makeUnstable) {
```
oleg-nenashev: It breaks the binary compatibility, a new constructor should be created

src/main/java/hudson/tasks/junit/JUnitTask.java
```diff
@@ -12,4 +12,6 @@
     boolean isKeepLongStdio();
 
     boolean isAllowEmptyResults();
+    
+    boolean isMakeUnstable();
```
oleg-nenashev: Breaks the binary compatibility of the interface.
In order to do so, you may need to update the plugin dependency to Jenkins Core 2.60.3 and Java 8. Then you will be able to add a default method to the interface

pom.xml
nguoianphu: Fixed.

src/main/java/hudson/tasks/junit/JUnitParser.java
nguoianphu: New constructor is created.

src/main/java/hudson/tasks/junit/JUnitTask.java
nguoianphu: Updated Java to 8. Jenkins core is 2.7.3 before.

nguoianphu:
Hi @oleg-nenashev 

Can you take a look if I need to update anything else, plz? I can't resolve the build error on ci.jenkins.io :(

amitSA:
+1 looking for this feature as my team has failing tests retried & reported, but we don't want the Junit plugin to set the build status. 

I know this PR is old, but many thanks if it can be added!

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -155,6 +174,7 @@ public void perform(Run build, FilePath workspace, Launcher launcher,
 
         if (action != null && action.getResult().getFailCount() > 0)
             build.setResult(Result.UNSTABLE);
+            listener.getLogger().println(Messages.JUnitResultArchiver_ChangeState("UNSTABLE"));
```
aabrahamian: I believe the conditional above will need braces otherwise this is considered outside of the block and will always execute, which may not have been your intent due to the indentation. If you did want it to always execute, you should update the indentation to reflect that.

nguoianphu: @aabrahamian Thank you. Corrected it.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -46,4 +46,7 @@ THE SOFTWARE.
     <f:entry title="${%Allow empty results}" field="allowEmptyResults">
         <f:checkbox default="false" title="${%Do not fail the build on empty test results}"/>
     </f:entry>
+    <f:entry title="${%Make the build UNSTABLE}" field="makeUnstable">
```
timja, line 49: this needs help text:
https://github.com/nguoianphu/junit-plugin/tree/makeUnstable/src/main/resources/hudson/tasks/junit/JUnitResultArchiver

nguoianphu: Will fix it. Thanks

RafaelloLoliop:
Oh I also need this one in my pipelines! I hope after 3 years it is still an option to finish this one!

emanuelelevo:
https://github.com/nguoianphu/junit-plugin/pull/1/files

timja:
> [nguoianphu/junit-plugin#1 (files)](https://github.com/nguoianphu/junit-plugin/pull/1/files)

you can also submit it directly to this repo

timja:
Still needs conflict resolution

PR #99, opened 2018-01-17
Fix a referenced issue number

chickenkiller:
The issue number for nested test suites is JENKINS-6545, not JENKINS-6454.


abayer:
Thanks!

PR #101, opened 2018-03-06
Use Run instead of AbstractBuild in AggregatedTestResultAction

ngg:
We're experimenting with aggregating test results from pipelines.
If the child is not an AbstractBuild then child.owner is null while child.run is filled, this fixes that case.


jglick:
FWIW I deliberately declined to touch `AggregatedTestResultAction` when adding Pipeline support to this plugin, since the typical use case was specific to freestyle job chains and solved more easily in Pipeline by simply running the `junit` step multiple times (perhaps in `parallel`).

PR #102, opened 2018-04-19
[DO NOT MERGE] JEP-305 demo

jglick:
See https://github.com/jenkinsci/plugin-pom/pull/100.


jglick:
Merely served as a demo of consuming Incrementals without also producing; not too interesting.

PR #103, opened 2018-04-27
Updated to new parent

jglick:
`Messages` is `@Restricted(NoExternalUse.class)`. We were just failing to enforce that before.


jglick:
Build failure in `master` seems to be a regression from @Wadeck’s commit, BTW—it enabled Windows-based builds, without fixing all tests to run on Windows.

PR #104, opened 2018-04-30
Avoid test to fail under Windows

Wadeck:
- after the addition of Windows in the desired type of node, some tests were failing due to some `sh` commands in pipeline

@reviewbybees


Wadeck:
duplicated by #100

PR #105, opened 2018-05-22
Workaournd to "[JENKINS-6048] Matrix builds fail when Axes values contain invalid characters for Windows directory names"

jeannich:
By default, for Matrix jobs, jenkins creates a workspace based on axis name and axis value.
In this test axis name is 'a', which is fine, but axis value is '<|#)' which cannot be used as folder name.
Overriding folder name to prevent this issue.

### DIFF
--- a/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
+++ b/src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
@@ -333,6 +333,7 @@ public class JUnitResultArchiverTest {
 
         MatrixProject p = j.jenkins.createProject(MatrixProject.class, "test-" + j.jenkins.getItems().size());
         p.setAxes(new AxisList(new TextAxis("a", "<|#)")));
+        p.setChildCustomWorkspace("axis_sub_folder");
         p.setScm(new SingleFileSCM("report.xml", getClass().getResource("junit-report-20090516.xml")));
         p.getPublishersList().add(new JUnitResultArchiver("report.xml"));
 


olivergondza:
I might be wrong but is not the test (provided it is failing right now) detecting an actual problem so it should be addressed instead of suppressing it in the test?

jeannich:
I am not sure of the purpose of this test but it doesn't seem to be targeting the issue JENKINS-6048.
I'll double check that.

With this workaround we could bring back tests execution on Windows!

jeannich:
The function tests "specialCharsInRelativePath" has been delivered by @olivergondza with this checkin https://github.com/jenkinsci/junit-plugin/commit/f81c822e2eed0ed0a495cb10a9482e9e8dee87a6

The purpose was to test *javascript* handling of special characters ; it has not link with workspace path issue.
Furthermore this workspace issue is already tracked by [JENKINS-6048](https://issues.jenkins-ci.org/browse/JENKINS-6048) hence I think this PR can be safely allowed

olivergondza:
Well, yes, the test was introduced to cover different scenario but I am concerned it will no longer be valid test it after this change.

So, let's first enable the test on windows so we see it failing ([it is being skipped currently](https://ci.jenkins.io/job/Plugins/job/junit-plugin/view/change-requests/job/PR-105/1/testReport/hudson.tasks.junit/JUnitResultArchiverTest/windows___Archive__windows____specialCharsInRelativePath/)) and then add this fix in next commit so we see it gets fixed.

jeannich:
No idea why but I did not have the line "Assume.assumeFalse(Functions.isWindows());" on my file!
Indeed @olivergondza with this restriction we are fine.

Just a remark though, maybe this line should be followed by a
"//TODO: remove this once JENKINS-6048 gets fixed"

I'll delete this PR (don't know how to do this. Feel free to go ahead and delete it)

PR #106, opened 2018-06-26
Add a list of all Skipped tests to the Test Result report

nirgilboa:
In a large test suite, it is useful to easily see a list of all skipped tests, as well as their age, i.e. for how many builds they've been skipped.
Previously it was only possible to view all tests in a given class, then sort by result to view the Skipped tests in it.

Note that I've left the "Duration" column, as tests may sometimes be skipped during execution, but it might be better to remove it to reduce clutter - I could use some feedback on that.

image

### DIFF
--- a/src/main/resources/hudson/tasks/test/MetaTabulatedResult/body.jelly
+++ b/src/main/resources/hudson/tasks/test/MetaTabulatedResult/body.jelly
@@ -25,6 +25,44 @@ THE SOFTWARE.
 
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson/test" xmlns:f="/lib/form">
+  <st:once>
+    <script type="text/javascript">
+      function showSkipped() {
+        const skippedTestsTable = document.getElementsByClassName("block collapse skipped")[0];
+        const showSkippedLink = document.getElementById("showSkippedLink");
+        skippedTestsTable.classList.add("show");
+        showSkippedLink.style.display = "none";
+      }
+    </script>
+    <style type="text/css">
+      *, ::after, ::before {
+      box-sizing: border-box;
+      }
+
+      .collapse {
+      display: block;
+      max-height: 0px;
+      overflow: hidden;
+      transition: max-height 0.5s cubic-bezier(0, 1, 0, 1);
+      }
+      .collapse.show {
+      max-height: 99em;
+      transition: max-height .5s ease-in-out;
+      }
+
+      .block {
+      margin-top: 10px;
+      padding: 0;
+      }
+      .block__content {
+      height: 100%;
+      }
+      a:link#showSkippedLink {
+      margin-top: 0.83em;
+      }
+    </style>
+  </st:once>
+
   <j:if test="${it.failCount!=0}">
     <h2>${%All Failed Tests}</h2>
     <table class="pane sortable bigtable stripped">
@@ -41,28 +79,6 @@ THE SOFTWARE.
           </td>
           <td class="pane" style="text-align:right;">
             <a href="${rootURL}/${f.failedSinceRun.url}" class="model-link inside">${f.age}</a>
-          </td>
-        </tr>
-      </j:forEach>
-    </table>
-  </j:if>
-
-  <j:if test="${it.skipCount!=0}">
-    <h2>${%All Skipped Tests}</h2>
-    <table class="pane sortable bigtable stripped">
-      <tr>
-        <td class="pane-header">${%Test Name}</td>
-        <td class="pane-header" style="width:4em">${%Duration}</td>
-        <td class="pane-header" style="width:3em">${%Age}</td>
-      </tr>
-      <j:forEach var="f" items="${it.skippedTests}" varStatus="i">
-        <tr>
-          <td class="pane no-wrap"><t:failed-test result="${f}" url="${f.getRelativePathFrom(it)}"/></td>
-          <td class="pane no-wrap" style="text-align:right;" data="${f.duration}">
-            ${f.durationString}
-          </td>
-          <td class="pane" style="text-align:right;">
-            <a href="${rootURL}/${f.skippedSinceRun.url}" class="model-link inside">${f.age}</a>
           </td>
         </tr>
       </j:forEach>
@@ -117,4 +133,34 @@ THE SOFTWARE.
       </tbody>
     </table>
   </j:if>
+
+  <j:if test="${it.skipCount!=0}">
+    <a id="showSkippedLink" name="editSkipped"
+       href="#showskipped"
+       onclick="javascript:showSkipped()">${%Show all skipped tests} ${">>>"}</a>
+    <div class="block collapse skipped">
+    <div class="block__content">
+    <h2>${%All Skipped Tests}</h2>
+    <table class="pane sortable bigtable stripped">
+      <tr>
+        <td class="pane-header">${%Test Name}</td>
+        <td class="pane-header" style="width:4em">${%Duration}</td>
+        <td class="pane-header" style="width:3em">${%Age}</td>
+      </tr>
+      <j:forEach var="f" items="${it.skippedTests}" varStatus="i">
+        <tr>
+          <td class="pane no-wrap"><t:failed-test result="${f}" url="${f.getRelativePathFrom(it)}"/></td>
+          <td class="pane no-wrap" style="text-align:right;" data="${f.duration}">
+            ${f.durationString}
+          </td>
+          <td class="pane" style="text-align:right;">
+            <a href="${rootURL}/${f.skippedSinceRun.url}" class="model-link inside">${f.age}</a>
+          </td>
+        </tr>
+      </j:forEach>
+    </table>
+    </div>
+    </div>
+  </j:if>
+
 </j:jelly>


nirgilboa:
@rsandell @uhafner  I'd appreciate your feedback, thanks :smiley:

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -407,16 +414,42 @@ else if (getRun() != null) {
...
     }
-    
+
+    /**
+     * If this test was skipped, then return the build number
+     * when this test was first skipped.
+     */
+    @Exported(visibility=9)
+    public int getSkippedSince() {
+        // If we haven't calculated failedSince yet, and we should,
```
uhafner: skippedSince

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -407,16 +414,42 @@ else if (getRun() != null) {
...
+        // do it now.
+        if (skippedSince==0 && getSkipCount()==1) {
+            CaseResult prev = getPreviousResult();
+            if(prev!=null && !prev.isPassed())
+                this.skippedSince = prev.getSkippedSince();
```
uhafner: What happens if in `previous` the test failed and therefore it has been disabled now? Then `previous.isPassed `is `false`, too. Might be a good setup for an integration test.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -407,16 +414,42 @@ else if (getRun() != null) {
...
+     * If this test was skipped, then return the build number
+     * when this test was first skipped.
+     */
+    @Exported(visibility=9)
+    public int getSkippedSince() {
+        // If we haven't calculated failedSince yet, and we should,
+        // do it now.
+        if (skippedSince==0 && getSkipCount()==1) {
+            CaseResult prev = getPreviousResult();
+            if(prev!=null && !prev.isPassed())
```
uhafner: It would make sense to always use `{}` for ifs.

nirgilboa:
Added an integration test which indeed helped to fix a logical error in getSkippedSince().
Thanks for your feedback!!

src/main/java/hudson/tasks/junit/CaseResult.java
nirgilboa: You're absolutely right - when a test becomes skipped, its skippedSince counter is set to the current build number.
In a subsequent build:
- If the test passes or fails, the skippedSince counter will be reset to 0.
- If the test is skipped again, the skippedSince is carried over from the previous build.

src/test/java/hudson/tasks/junit/HistoryTest.java
```diff
@@ -108,7 +108,87 @@ public void testFailedSince() throws Exception {
...
+        assertNotNull(eleanorCase);
         assertTrue("eleanor failed", !eleanorCase.isPassed());
-        assertEquals("eleanor has failed since build 3", 3, eleanorCase.getFailedSince()); 
+        assertEquals("eleanor has failed since build 3", 3, eleanorCase.getFailedSince());
+    }
+
+    @LocalData
+    @Test
+    public void testSkippedSince() throws Exception {
+        assertNotNull("project should exist", project);
```
uhafner: It is quite hard to understand if the setup is not visible and packed into a zip file. (But it is not your fault, as this approach has been used in the existing test case as well.)

nirgilboa: Completely agree :) it took running the test and examining its log in order to understand that a zip file was used..

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -96,6 +96,13 @@
      */
     private /*final*/ int failedSince;
 
+    /**
+     * This test has been skipped since this build number (not id.)
+     *
+     * If {@link #isPassed() passing}, this field is left unused to 0.
+     */
+    private /*final*/ int skippedSince;
```
deepy: shouldn't `/*final*/` be removed?

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -397,26 +404,52 @@ public int getPassCount() {
...
         }
         return failedSince;
     }
-    
+
+    /**
+     * If this test was skipped, then return the build number
+     * when this test was first skipped.
+     */
+    @Exported(visibility=9)
```
deepy: Shouldn't the visibility of the skipped tests be higher than the visibility of failed tests?

src/main/java/hudson/tasks/junit/CaseResult.java
nirgilboa: Thanks - removing

src/main/java/hudson/tasks/junit/CaseResult.java
nirgilboa: On second thought I think the visibility for getFailedSince and getSkippedSince should be identical.

nirgilboa:
@jglick ping

nirgilboa:
@jglick @abayer 
Could you please review and decide on this?

src/main/java/hudson/tasks/test/MetaTabulatedResult.java
```diff
@@ -41,4 +41,9 @@
      */
     public abstract Collection<? extends TestResult> getFailedTests();
 
+    /**
+     * All skipped tests.
+     */
+    public abstract Collection<? extends TestResult> getSkippedTests();
```
olivergondza: Careful, this is going to blow up for [implementations](https://github.com/search?q=org%3Ajenkinsci+%22extends+MetaTabulatedResult%22&type=Code) that predate this change as they will not implement this method. Must be non-abstract.

src/main/resources/lib/hudson/test/non-passed-test.jelly
```diff
@@ -25,19 +25,19 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
   <st:documentation>
-    Display link to the failed test.
+    Display link to the test.
```
olivergondza: Ditto, for renaming a reusable tag. There appears to be at least one[1] external use of this.

[1] https://github.com/jenkinsci/mentor-questa-vrm-plugin/blob/04079b04117d351878e1be16d19c79fa38ed2568/src/main/resources/com/mentor/questa/vrm/jenkins/QuestaVrmRegressionResult/body.jelly#L45

src/main/resources/lib/hudson/test/aggregated-failed-tests.jelly
```diff
@@ -52,7 +52,7 @@ THE SOFTWARE.
           </tr>
           <j:forEach var="f" items="${report.result.failedTests}" varStatus="i">
             <tr>
-              <td class="pane no-wrap"><t:failed-test result="${f}" url="../${report.child.project.shortUrl}testReport/${f.getRelativePathFrom(report.result)}"/></td>
+              <td class="pane no-wrap"><t:non-passed-test result="${f}" url="../${report.child.project.shortUrl}testReport/${f.getRelativePathFrom(report.result)}"/></td>
```
olivergondza: Ditto, for renaming a reusable tag. There appears to be at least one[1] external use of this.

[1] https://github.com/jenkinsci/mentor-questa-vrm-plugin/blob/04079b04117d351878e1be16d19c79fa38ed2568/src/main/resources/com/mentor/questa/vrm/jenkins/QuestaVrmRegressionResult/body.jelly#L45

olivergondza:
I like the change in general. One small thing I would like to see improved is making the list of skipped tests *less* significant in the UI. Collapsed by default? Moved to the very bottom? It is useful for sure but probably distracting most of the time...

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -397,26 +404,52 @@ public int getPassCount() {
...
+    public int getSkippedSince() {
+        // If we haven't calculated skippedSince yet, and we should, do it now.
+        if (skippedSince==0 && getSkipCount()==1) {
+            CaseResult prev = getPreviousResult();
+            if(prev!=null && prev.isSkipped()){
+                this.skippedSince = prev.getSkippedSince();
```
zbynek: Unlike failedSince, skippedSince is not available in XML, so this could recursively load all the skipped builds leading to StackOverflowError like [JENKINS-31660](https://issues.jenkins-ci.org/browse/JENKINS-31660) . IMHO it would be safer to save skippedSince in XML too and assume current build is the first skipped one if no information is found in XML, avoiding recursion. (Related to #117.)

nirgilboa: Good catch! I will review this and suggest a fix

nirgilboa: I've given this some thought.. as I understand we only parse the XML and have no control over its contents. Do you see any other way to protect against a StackOverflowError ?

nirgilboa:
Skipped tests table collapsed by default, accessible using a link placed at the bottom of the page.
Screenshots below:

Screenshot_ubuntu17 10_2019-05-31_16:48:42

Link is hidden and table is shown:
Screenshot_ubuntu17 10_2019-05-31_16:49:08

zbynek:
The skippedSince value is saved to XML (automatic saving of all fields, IIUC implemented by XStream2), but in most cases it's not computed before saving. See `SuiteResultTest::testSuiteResultPersistence` for implementation of saving test. For `failedSince` the freeze method makes sure that the field is computed before saving. If you add `recomputeSkippedSinceIfNeeded()` to `CaseResult::freeze`, saving will work, but it may lead to loading all the previous results from the disk. So `recomputeSkippedSinceIfNeeded` should fall back to previous build number if previous build failed but we don't know since when it's failing.

Also fixing the freeze method will cause a failure of the test `testPreviousBuildNotLoaded` from #117. Other way of testing for deep recursion is needed (maybe include 2 builds and make sure that we don't call `getPreviousBuild` on the older one?)

hodnajit:
Is it possible, that this feature will be merged? :) I would like to use this feature as well :)

timja:
> I'd love to see this feature available. What is lacking for a release? Basically lack of engagement? :-|

reviewing if https://github.com/jenkinsci/junit-plugin/pull/106#issuecomment-497928233 needs fixing and then resolving conflicts and testing it end 2 end. Basically needs someone (anyone) to pick it up and get it over the finish line.

nirgilboa:
I'll be happy to revive this and push it if there's demand :)

SebastianSolidwork:
Nice! There is demand :-) How could you have waited for so long !?!?12 ;-)

I would interested in the All Tests the least.
How about making All Skipped Tests as second table, like in the original comment?
If this is too much for here, I would create another issue.

olivergondza:
@nirgilboa, please resolve the conflicts. Thanks!

SebastianSolidwork:
@nirgilboa Any news on this one?

PR #109, opened 2018-08-20
TestResultParser was never a meaningful extension point

jglick:
7a6dcfe5838aea0f306eb537de3039fa7ef6ee1e claimed to be introducing an “extension point” with a particular behavior—that you could register a new one in a plugin and it would be called by the standard publisher task. This does not appear to have ever been true, as you can [see here](https://github.com/jenkinsci/junit-plugin/blob/7a6dcfe5838aea0f306eb537de3039fa7ef6ee1e/core/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java#L108). Similarly [here](https://github.com/jenkinsci/cucumber-testresult-plugin/blob/485b03b5ba8730ffa7cf9057f31e7d4f29fcc36f/src/main/java/org/jenkinsci/plugins/cucumber/jsontestsupport/CucumberTestResultArchiver.java#L132), and in `utplsql` and `mentor-questa-vrm` (see [index](https://jenkins.io/doc/developer/extensions/junit/#testresultparser)), and `ecutest` (unindexed). The aggregated commit from Subversion even includes the note that

> The all() method is removed after the following discussion:
>>> - I propose we remove AbstractTestResult.all(). I don't think being able
>>> to enumerate AbstractTestResult subtypes buy us anything, since those
>>> can only be instantiated by parsers that intimately know how to
>>> instantiate them properly. Or did I miss the purpose of this?
>>
>> I put in the all method because it was suggested by the wiki:
>> "You also need to define the static "all" method..."
>> http://wiki.hudson-ci.org/display/HUDSON/Defining+a+new+extension+point
>> so I don't have a problem with removing it.

yet the `all()` method remained.

In the interests of removing dead code and reducing confusion I am just deprecating everything that does not seem to actually be used. I am retaining the fact that `JUnitParser extends TestResultParser` to avoid breaking callers that might be linking against that binary signature, like [`PipelineRealtimeTestResultAction.parse`](https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/blob/561b341c94890393fe494f63c5c2becaa251ceae/src/main/java/org/jenkinsci/plugins/junitrealtimetestreporter/PipelineRealtimeTestResultAction.java#L74), and retaining the `@Extension` to avoid breaking `labeled-test-groups-publisher`.

### DIFF
--- a/src/main/java/hudson/tasks/test/TestResultParser.java
+++ b/src/main/java/hudson/tasks/test/TestResultParser.java
@@ -25,6 +25,7 @@ package hudson.tasks.test;
 
 import hudson.AbortException;
 import hudson.ExtensionList;
+import hudson.ExtensionPoint;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
@@ -37,7 +38,9 @@ import javax.annotation.Nonnull;
 
 /**
  * Parses test result files and builds in-memory representation of it as {@link TestResult}.
- *
+ * <p>
+ * Originally this was treated as an {@link ExtensionPoint},
+ * but no supported code path actually uses this API directly. See {@link #all}.
  * <p>
  * Parsers are stateless, and the {@link #parseResult} method
  * can be concurrently invoked by multiple threads for different builds.
@@ -49,7 +52,7 @@ public abstract class TestResultParser {
      * Returns a human readable name of the parser, like "JUnit Parser".
      *
      * @return a human readable name of the parser, like "JUnit Parser".
-     * @deprecated Unused.
+     * @deprecated Normally unused.
      */
     @Deprecated
     public String getDisplayName() {
@@ -61,7 +64,7 @@ public abstract class TestResultParser {
      * For example, "JUnit XML reports:"
      *
      * @return the text is used in the UI prompt for the GLOB that specifies files to be parsed by this parser.
-     * @deprecated Unused.
+     * @deprecated Normally unused.
      */
     @Deprecated
     public String getTestResultLocationMessage() {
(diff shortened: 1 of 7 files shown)


oleg-nenashev:
https://github.com/jenkinsci/labeled-test-groups-publisher-plugin/blob/d2130260c531756b85afdfc696c05d541f6b5e65/src/main/java/hudson/plugins/labeledgroupedtests/LabeledTestResultGroupPublisher.java#L78-L91 seems to be doing something reasonable with parsers. It is used in order to offer users a selection of parsers in WebUI: https://github.com/jenkinsci/labeled-test-groups-publisher-plugin/blob/master/src/main/resources/hudson/plugins/labeledgroupedtests/LabeledTestGroupConfiguration/config.jelly#L26

Such use-cases will be broken by the change

jglick:
With only 231 installations, and functionality which is completely superseded by a supported feature (`junit` within Pipeline `stage`s), I would not be sad to break that plugin. Still, this could be reworked to remain compatible while making it clear that this code path is not used in any supported scenario.

src/main/java/hudson/tasks/test/TestResultParser.java
```diff
@@ -32,37 +32,29 @@
...
+ * but no supported code path actually uses this API directly. See {@link #all}.
  * <p>
  * Parsers are stateless, and the {@link #parseResult} method
  * can be concurrently invoked by multiple threads for different builds.
  *
  * @since 1.343
- * @see DefaultTestResultParserImpl
  */
-public abstract class TestResultParser implements ExtensionPoint {
+public abstract class TestResultParser {
```
oleg-nenashev: It won't be compatible in all cases after this change. Not sure what is the need in this change

jglick: `ExtensionPoint` has no methods, and is not even treated as a marker interface at runtime. It is used purely for documentation, including [this list](https://jenkins.io/doc/developer/extensions/). Since `TestResultParser` is not in fact being used as an extension point, and AFAICT never was, this clears up the point of confusion.

PR #110, opened 2018-08-28
External storage prototype

jglick:
Prototype of an SPI for keeping test results anywhere other than `junitResults.xml`.

Subsumes #109.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools.incrementals</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-beta-5</version>
+  </extension>
+</extensions>
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,2 @@
+-Pconsume-incrementals
+-Pmight-produce-incrementals
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.60.1'])
+buildPlugin()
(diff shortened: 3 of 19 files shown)


timja:
Hey @jglick 

What would it take to get some movement on this?
Did you implement a prototype consumer of this api?

I'm happy to spend a bit of time on this if you think it could be moved forward

jglick:
@timja indeed this was put on hold due to lack of interest from my employer in spending resources on [this general area](https://jenkins.io/sigs/cloud-native/pluggable-storage/). You should be able to get a rough idea of how far I got by looking at outstanding `TODO` comments and the coverage in `TestResultStorageTest`. If I get permission to spend some more time here, I would want to get it far enough along that all the basic functions of the plugin work (so far as automated tests are concerned) against the dummy provider using local H2 SQL, including at least one complex feature such as the trend graph implemented using a proper query. That would be enough for someone else with more practical database / cloud storage experience to prototype a realistic provider plugin (not necessarily SQL-based), which could be the reference implementation for a JEP, and help fill in some of the gaps in functionality needed for production use.

What is your interest in this? Is there a particular issue that is hurting you in the field—disk space consumption after many builds, difficulty creating custom reports, compliance requirement for archival storage, Remoting channel congestion transferring massive test results, slow build load time parsing `junitResult.xml`, etc.?

timja:
Specific interest in this currently is:
We're trying to decouple our jenkins from our infrastructure and reduce the access that it has, it's basically a god everywhere right now.

In this case we're running our tests in Kubernetes and wanting to report them back to jenkins, not have jenkins collect them

We were easily able to get artifacts back to jenkins using: https://github.com/jenkinsci/azure-artifact-manager-plugin and storing them in blob storage. It would be nice to be able to do something similar with the test results

General interest:
Jenkins is our most critical piece of infrastructure, anything that moves its storage off of the master and into blob storage is great for us.
Makes it more resilient to failure and easier to upgrade.

We also have disk spaces issues, 
image

I suspect the test reports don't contribute much to our disk space problems, but its hard to tell as there's soo many files that `du` doesn't respond in a reasonable time, I can run it against one teams jobs folder, come back 10 minutes later and not have a result.

timja:
Replaced by https://github.com/jenkinsci/junit-plugin/pull/141

src/main/java/hudson/tasks/test/DefaultTestResultParserImpl.java
```diff
@@ -48,7 +48,9 @@
  *
  * @since 1.343
  * @author Kohsuke Kawaguchi
+ * @deprecated Unused, unusable.
  */
+@Deprecated
```
jtnord, line 53: https://github.com/jenkinsci/cucumber-testresult-plugin/blob/master/src/main/java/org/jenkinsci/plugins/cucumber/jsontestsupport/CucumberJSONParser.java#L45

PR #116, opened 2019-04-11
Avoid logging useless messages at INFO

jglick:
After running a hello-world Pipeline with some tests, I noticed junk in the system log:

```
2019-04-11 14:14:01.146+0000 [id=15]	INFO	hudson.tasks.test.TestObject#getRelativePathFrom: trying to get relative path, but it is not my ancestor, and we're not in a stapler request. Trying absolute hudson url...
2019-04-11 14:14:01.148+0000 [id=15]	INFO	hudson.tasks.test.TestObject#getRelativePathFrom: Here's our relative path: http://jenkins//job/sample-pipeline/2//testReport/junit/test/SomeTest/test6/
```

Seems to be very old code, not a recent addition. Note the incorrect double slashes in two spots in the URL path—harmless?

### DIFF
--- a/src/main/java/hudson/tasks/test/TestObject.java
+++ b/src/main/java/hudson/tasks/test/TestObject.java
@@ -45,6 +45,7 @@ import org.kohsuke.stapler.interceptor.RequirePOST;
 import javax.servlet.ServletException;
 import java.io.IOException;
 import java.util.*;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/tasks/test/TestObject.java
```diff
@@ -199,7 +200,7 @@ public String getRelativePathFrom(TestObject it) {
                 buf.insert(0, hudsonRootUrl);
             }
 
-            LOGGER.info("Here's our relative path: " + buf.toString()); 
+            LOGGER.log(Level.FINE, "Here is our relative path: {0}", buf);
```
jvz:
```suggestion
            LOGGER.fine(() -> "Here is our relative path: " + buf);
```

jglick: You can use the lambda form too. In my tests it is slightly more expensive for one argument, slightly less for two or more. The bytecode is a bit longer. :man_shrugging:

PR #117, opened 2019-04-13
[JENKINS-31660] avoid reading unnecessary test results from disk

zbynek:
See [JENKINS-31660](https://issues.jenkins-ci.org/browse/JENKINS-31660)

In case the a test was skipped, we don't need to know if it had previous failures, which allows us to avoid loading a lot of data from the disk and avoid StackOverflowError (in case the test was always skipped, we avoid loading all the previous builds).

There are two methods that recompute the failedSince field, one of them handled skipped builds correctly, the other one did not. With this PR they will both use the same logic.


zbynek:
From a few days of testing on a server with 10k+ builds it seems the StackOverflowErrors are gone and the server is more responsive. @olivergondza I would like to ask you or one of the other maintainers for a review.

zbynek:
@jenkinsci/code-reviewers sorry for nagging, but could someone please have a look at this?

PR #118, opened 2019-04-19
[JENKINS-39203] Mark JUnit steps with failing tests with WarningAction

dwnusbaum:
See https://github.com/jenkinsci/workflow-api-plugin/pull/91 and [JENKINS-39203](https://issues.jenkins-ci.org/browse/JENKINS-39203). (I am intentionally not updating the status of that ticket or assigning it to myself until I am ready to merge/release all relevant changes to avoid giving users premature hope of a fix and so that I have clear answers for what will and won't work and when any fixes will be released.) Subsumes and is based on #96.

This PR adds a `WarningAction` to the `StepAtomNode` for `junit` steps with failing tests so that the actual stage containing the `junit` step can be visualized as unstable via https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/25.

Desired reviewers: @abayer, @jglick, @car-roll

### DIFF
--- a/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
+++ b/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
@@ -6,13 +6,13 @@ import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.junit.JUnitResultArchiver;
+import hudson.tasks.junit.TestResult;
 import hudson.tasks.junit.TestResultAction;
 import hudson.tasks.junit.TestResultSummary;
 import hudson.tasks.test.PipelineTestDetails;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nonnull;
-
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
 import org.jenkinsci.plugins.workflow.actions.WarningAction;
@@ -52,12 +52,13 @@ public class JUnitResultsStepExecution extends SynchronousNonBlockingStepExecuti
             TestResultAction testResultAction = JUnitResultArchiver.parseAndAttach(step, pipelineTestDetails, run, workspace, launcher, listener);
 
             if (testResultAction != null) {
-                int testFailures = testResultAction.getResult().getResultByNode(nodeId).getFailCount();
+                TestResult testResult = testResultAction.getResult().getResultByNode(nodeId);
+                int testFailures = testResult.getFailCount();
                 if (testFailures > 0) {
                     node.addOrReplaceAction(new WarningAction(testFailures + " tests failed"));
                     run.setResult(Result.UNSTABLE);
                 }
-                return new TestResultSummary(testResultAction.getResult().getResultByNode(nodeId));
+                return new TestResultSummary(testResult);
             }
         } catch (Exception e) {
             listener.getLogger().println(e.getMessage());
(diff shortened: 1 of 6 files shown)


dwnusbaum:
CI failure is because my Jenkinsfile changes are being ignored and the builds on 2.60.x are failing because plugins aren't able to load correctly.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -50,8 +52,9 @@ protected TestResultSummary run() throws Exception {
             TestResultAction testResultAction = JUnitResultArchiver.parseAndAttach(step, pipelineTestDetails, run, workspace, launcher, listener);
 
             if (testResultAction != null) {
-                // TODO: Once JENKINS-43995 lands, update this to set the step status instead of the entire build.
-                if (testResultAction.getResult().getFailCount() > 0) {
+                int testFailures = testResultAction.getResult().getResultByNode(nodeId).getFailCount();
+                if (testFailures > 0) {
+                    node.addOrReplaceAction(new WarningAction(testFailures + " tests failed"));
                     run.setResult(Result.UNSTABLE);
```
jglick: Another option would be for `WarningAction.onAttached` to call `Run.setResult` automatically. On the other hand I can imagine wanting this drastic effect to be explicit.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -50,8 +52,9 @@ protected TestResultSummary run() throws Exception {
             TestResultAction testResultAction = JUnitResultArchiver.parseAndAttach(step, pipelineTestDetails, run, workspace, launcher, listener);
 
             if (testResultAction != null) {
-                // TODO: Once JENKINS-43995 lands, update this to set the step status instead of the entire build.
-                if (testResultAction.getResult().getFailCount() > 0) {
+                int testFailures = testResultAction.getResult().getResultByNode(nodeId).getFailCount();
```
jglick: I guess the insertion of the `getResultByNode` call here is intentional.

It means you can introduce a variable and use it on line 60.

dwnusbaum: Good catch, done in 030f0db.

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
dwnusbaum: I wasn't familiar with `onAttached` previously, but I think you are referring to [`RunAction2.onAttached`](https://github.com/jenkinsci/jenkins/blob/7936d873a1550b060acc25de0c910aac0fcbebf7/core/src/main/java/jenkins/model/RunAction2.java#L40). We certainly could make some changes in `FlowNode` so that `onAttached` and `onLoad` would be called for actions implementing `RunAction2`, but as you mentioned I think this particular effect is best left explicit.

jglick:
> my Jenkinsfile changes are being ignored

Ask to become a maintainer.

jglick:
Looks like you have some POM problems.

```
Structs Plugin v1.17 failed to load.
 - You must update Jenkins from v2.60.1 to v2.60.3 or later to run this plugin.
```

```
java.lang.NoSuchMethodError: jenkins.model.Jenkins.get()Ljenkins/model/Jenkins;
	at org.jenkinsci.plugins.workflow.cps.nodes.StepDescriptorCache.getPublicCache(StepDescriptorCache.java:48)
```

dwnusbaum:
No, just Jenkinsfile problems (https://github.com/jenkinsci/junit-plugin/pull/118/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3L1). I'll request to be a maintainer so my Jenkinsfile changes are used.

PR #120, opened 2019-06-27
JENKINS-48583: Improve logic that handles suite aggregation

khneal:
Original intended use case was for complete duplicate suite definitions due to aggregation of multiple suites into a single file by ant's JUnitReport task. I added a test to ensure that scenario is still supported (who still uses ant? :D ).

Modern tools run tests in parallel, and a single suite may have test cases written across several files. The updated logic will only ignore a suite if it has an identical name, timestamp, duration, and number of test cases as the existing. If both suites being compared only have a single test case, then the full name of test case will also be compared.

Also, any ignored test suites are now logged as a warning. Many productive hours have been lost investigating this issue, so hopefully that will minimize future hours lost if another use case is discovered.

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResult.java
+++ b/src/main/java/hudson/tasks/junit/TestResult.java
@@ -296,13 +296,22 @@ public final class TestResult extends MetaTabulatedResult {
                 // However, a common problem is that people parse TEST-*.xml as well as TESTS-TestSuite.xml.
                 // In that case consider the result file as a duplicate and discard it.
                 // see http://jenkins.361315.n4.nabble.com/Problem-with-duplicate-build-execution-td371616.html for discussion.
-                if(strictEq(s.getTimestamp(),sr.getTimestamp())) {
-                    String message = "Ignoring duplicate TestSuite: " + sr.getName();
-                    if (null != sr.getFile()) {
-                        message = message + " (" + sr.getFile() + ")";
+                if(strictEq(s.getTimestamp(),sr.getTimestamp()) &&
+                        s.getDuration() == sr.getDuration() &&
+                        s.getCases().size() == sr.getCases().size()) {
+                    
+                    // Rare edge case: testing framework writes one file per test case, but all same test suite...
+                    // *and* duration is not included, or so fast that first two files have 0.000 or 0.001.
+                    // When the second file is read, number of test cases will match... so compare the name.
+                    if (s.getCases().size() != 1 ||
+                            s.getCases().get(0).getFullName().equals(sr.getCases().get(0).getFullName())) {
+                        String message = "Ignoring duplicate TestSuite: " + sr.getName();
+                        if (null != sr.getFile()) {
+                            message = message + " (" + sr.getFile() + ")";
+                        }
+                        LOGGER.warning(message);
+                        return;
                     }
-                    LOGGER.warning(message);
-                    return;
                 }
 
                 duration += sr.getDuration();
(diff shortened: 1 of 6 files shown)


khneal:
Edited commit messages to add square brackets around JIRA issue ID

khneal:
FYI - it appears #20 was opened years ago to fix this same issue.

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -286,16 +288,30 @@ private void add(SuiteResult sr) {
...
-                    return;
+                if(strictEq(s.getTimestamp(),sr.getTimestamp()) &&
+                        s.getDuration() == sr.getDuration() &&
+                        s.getCases().size() == sr.getCases().size()) {
+                    
+                    // Rare edge case: testing framework writes one file per test case, but all same test suite...
+                    // *and* duration is not included, or so fast that first two files have 0.000 or 0.001.
+                    // When the second file is read, number of test cases will match... so compare the name.
+                    if (s.getCases().size() != 1 ||
+                            s.getCases().get(0).getFullName().equals(sr.getCases().get(0).getFullName())) {
```
jonlil: I'm no java developer but this seems to be very complex. Perhaps extract this logic to a private method and give it a self explanatory method name? That would also make the logic easier to test.

Regard the implementation I'm not sure adding comparisons for duration and filename will make this problem go away. I think there is two paths to take here.

1. calculate checksums for the report files and compare them
2. disable this logic with a configuration parameter like

**Jenkinsfile**
```groovy
junit 'tmp/junit.xml', disable_duplicate_report_check: true
```

I would like to see option 1 since that actually solves this bug. 
This implementation just makes it even harder to understand and debug.

caweidmann:
Any update on this when there will be a fix in? We have this issue in our pipeline and I spent the whole day yesterday trying to debug why we are getting inconsistent unit test results with Junit. Alternatively is there a better plugin to use with Jenkins to show coverage?

jonlil:
A workaround for this that we have been using for a long time in many projects now is the following

```bash
sed -i 's/timestamp="[^"]*"/timestamp="'"${TEST_SPLIT_INDEX}"'"/g' \
    ".jenkins/tmp/reports/junit.xml"
```

cc @caweidmann

src/main/java/hudson/tasks/junit/TestResult.java
khneal: > 1. calculate checksums for the report files and compare them
This would break the original use case from years ago, which was separate files per testsuite, then a single file with all testsuites aggregated. The checksum wouldn't match, but the intention was to not add duplicate tests.

> 2. disable this logic with a configuration parameter like
I think more and more people will run into false positives where tests are disappearing and it's difficult to figure out why. If the logic does not work correctly in the default configuration, then maybe it should be removed... which is actually the direction that I've started moving towards. See comment that I will post shortly.

khneal:
Based on all the people that have posted about this issue (in this PR as well as this and other JIRA issues), the deduplication logic is doing far more damage than good. The missing tests have already wasted a lot of person-hours, whereas I would argue the number of hours saved by the cases that have been correctly deduplicated is negligible. If this logic were removed, then the fix for removing duplicate test cases is... just don't include them in the file pattern. Simple. Or delete the aggregated file before the build finishes. Either way, waaay better than randomly missing some tests.

khneal:
I created PR #131 to remove the deduplication logic, so I'm closing this PR.
@jonlil - would you be willing to review that PR?

PR #123, opened 2019-08-16
Fix chart with deleted builds

il--ya:
When builds are deleted, there *are* gaps in build numbers, and chart doesn't show anything before deleted build. This causes [JENKINS-52711](https://issues.jenkins-ci.org/browse/JENKINS-52711).
I haven't tested this fix, as I am stuck at the moment with out-of-date Jenkins version, but I think it should work.

### DIFF
--- a/src/main/java/hudson/tasks/test/AbstractTestResultAction.java
+++ b/src/main/java/hudson/tasks/test/AbstractTestResultAction.java
@@ -222,7 +222,7 @@ public abstract class AbstractTestResultAction<T extends AbstractTestResultActio
             loadedBuilds = null;
         }
         while(true) {
-            b = loadedBuilds == null || loadedBuilds.contains(b.number - /* assuming there are no gaps */1) ? b.getPreviousBuild() : null;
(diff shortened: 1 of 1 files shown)


il--ya:
ping @jglick @dwnusbaum

il--ya:
I don't think tests update is necessary for such a simple bug fix. And I don't have capacity to do that anyway. @jglick do you see any particular reason for concern with this change?

zbynek:
@il--ya are you sure this won't load additional builds? Like if loaded builds contains 1 and 999, this may load builds 998, 997, ... 

Even though I'm not sure how realistic that scenario is, it may be safer to put loaded builds to a TreeSet and get the previous build as `loadedBuildsTreeSet == null ? current - 1 : loadedBuildsTreeSet.lower(current)`

jglick:
> do you see any particular reason for concern with this change?

Having written the condition being patched, yes I do. It is not obvious to me that this is right, especially if you say you have not tested it interactively either. This is a very widely used plugin, so any regression would likely be worse than the reported limitation. Would need to spend some time refamiliarizing myself with this.

PR #124, opened 2019-08-22
Fix chart with deleted builds

il--ya:
[JENKINS-25340] [JENKINS-52711]
When there are deleted builds, test results chart only shows results for until the deleted build, older builds are not displayed. This PR tries to address this.


zbynek:
For the record this is a replacement of #123 

@jglick's request for adding automated test is still unresolved

PR #126, opened 2019-10-23
Cleanup old code

res0nance:
- Bunch of minor cleanups


src/main/java/hudson/tasks/junit/History.java
```diff
@@ -69,14 +69,14 @@ public TestObject getTestObject() {
 	}
 	
     public boolean historyAvailable() {
-       if (testObject.getRun().getParent().getBuilds().size() > 1)
-           return true;
+       if (testObject.getRun().getParent().getBuilds().isEmpty())
```
zbynek: This is not equivalent to the old code (was size >1, now it's size >0)

src/main/java/hudson/tasks/test/TestObject.java
```diff
@@ -189,15 +204,15 @@ public String getRelativePathFrom(TestObject it) {
                 buf.insert(0, Functions.getRelativeLinkTo(myBuildAsItem));
             } else {
                 // We're not in a stapler request. Okay, give up.
-                LOGGER.fine("trying to get relative path, but it is not my ancestor, and we are not in a stapler request. Trying absolute hudson url...");
-                String hudsonRootUrl = Helper.getActiveInstance().getRootUrl();
-                if (hudsonRootUrl==null||hudsonRootUrl.length()==0) {
+                LOGGER.fine(() -> "trying to get relative path, but it is not my ancestor, and we are not in a stapler request. Trying absolute jenkins url...");
```
olivergondza: What is the point of lazily evaluating a string literal?

res0nance: Fixed it didn't realize it was just a pure string literal

PR #128, opened 2020-03-23
Add failed tests list to TestResultSummary

LosBandolero:
This feature is need in the pipelines, so for example we are notifying in slaack about out failed tests and by this feature we will be able to have a list directly in slack

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResultSummary.java
+++ b/src/main/java/hudson/tasks/junit/TestResultSummary.java
@@ -12,6 +12,7 @@ public class TestResultSummary implements Serializable {
     private int skipCount;
     private int passCount;
     private int totalCount;
+    private List<CaseResult> failedTests;
 
     public TestResultSummary() {
     }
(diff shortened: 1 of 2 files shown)


LosBandolero:
@oleg-nenashev

src/main/java/hudson/tasks/junit/TestResultSummary.java
```diff
@@ -12,6 +13,7 @@
     private int skipCount;
     private int passCount;
     private int totalCount;
+    private List<CaseResult> failedTests;
```
oleg-nenashev: This class does not seem to be serializable

src/main/java/hudson/tasks/junit/TestResultSummary.java
```diff
@@ -21,6 +23,7 @@ public TestResultSummary(TestResult result) {
         this.skipCount = result.getSkipCount();
         this.passCount = result.getPassCount();
         this.totalCount = result.getTotalCount();
+        this.failedTests = result.getFailedTests();
```
oleg-nenashev: Maybe makes sense to double check that the collection is serializable, e.g. via copy constructor

varikin:
Any update on this PR? We also want this feature.

varikin:
Thank you. I opened https://github.com/jenkinsci/junit-plugin/issues/410

PR #129, opened 2020-03-25
Add support and access to junit properties

thomas-dee:
It would be great if the testsuite/properties tags would be accessible.


timja:
Needs a unit test, some more expansion on the use case would be useful

apopescu99:
Hello, will this change be included in the junit-plugin soon? Thank you!

timja:
not without the review comment being addressed

mPokornyETM:
> It would be great if the testsuite/properties tags would be accessible.

It will be great when you show this properties somewhere in test-suite page

timja:
Added in https://github.com/jenkinsci/junit-plugin/pull/541

PR #131, opened 2020-04-07
JENKINS-48583 Remove test result deduplication logic

khneal:
The most important change in this PR removes de-duplication of test suites based on time stamp:
```
// However, a common problem is that people parse TEST-*.xml as well as TESTS-TestSuite.xml.
// In that case consider the result file as a duplicate and discard it.
// see http://jenkins.361315.n4.nabble.com/Problem-with-duplicate-build-execution-td371616.html for discussion.
if(strictEq(s.getTimestamp(),sr.getTimestamp())) {
    return;
}
```
From http://jenkins.361315.n4.nabble.com/Problem-with-duplicate-build-execution-td371616.html, @kohsuke said:
> I think the best way to resolve this is to make Hudson smarter --- the
<testsuite> element seems to have the timestamp, so Hudson should be
able to use that and determine that two <testsuite>s with the same name
and timestamp are identical.

This may have been true at the time, but nowadays it is not uncommon to launch test cases in parallel. This means the same test suite may be spread across multiple files, many of which have an identical timestamp down to the millisecond, and that many customers are encountering "missing" tests from the UI that exist in a valid format in the files. One option would be to pile on logic to narrow the scope of deduplication... I tried that in PR #120, but my organization (Intuit) found additional cases that were incorrectly excluded. I now believe that fixing the logic would unnecessarily complicate the code for very little benefit. The value received by auto-de-duplicating test suites (which the end user could also accomplish by updating the junit result file spec or just deleting the duplicate aggregation file before the post build step) pales in comparison to the cost of missing valid tests (as shown by the several JIRA issues filed for this problem and all the comments on PRs trying to fix it).

On top of that, the first commit for this PR shows that the original use case where Ant duplicates test suites into an aggregated file has not been supported for many released versions of the junit-plugin (the test added in the first commit should pass, but does not and fails the build). While I included a fix for that use case in my second commit, I still don't believe the de-duplication should be supported.

I think there's a strong case for this change. Let me know if there's anything else I can do to advocate for getting this PR merged. Thank you for reading!


khneal:
Side note: the force-push was simply to rebuild the commit that previously failed, possibly because Jenkins was restarted: 
> `09:30:53  Resuming build at Wed Apr 08 16:30:53 UTC 2020 after Jenkins restart
`
> ...
> `09:30:54  java.lang.IllegalStateException: JENKINS-50407: no loaded shell in CpsFlowExecution[Owner[Plugins/junit-plugin/PR-131/2:Plugins/junit-plugin/PR-131 #2]]`

khneal:
Sorry for all the force pushes. I was trying to have the commits tell a chronological story.
Also, several of the Jenkins builds failed for reasons beyond my control, and a force push was the only way I know that I'm able to re-trigger a build.
This PR is ready for feedback now.

khneal:
I'm still interested in moving this forward, but I can understand there are a lot of things up in the air right now... so I'll check back in a week or two.

khneal:
I’m still watching this for comments. Not trying to be impatient... just showing continued commitment.

khneal:
Just checking in to see if anybody has seen this. Feel free to thumbs up this comment.

khneal:
Hello, it's been a month since I opened this PR and I still don't see any comments (but shout out to amai2012 for the thumbs up!). If there's something I can do to help this move forward, please let me know.

khneal:
@jglick - do you have any recommendations for me?

jglick:
Sorry, I am not a maintainer.

khneal:
@jglick - thank you for the reply. Interestingly, your name is the only one listed under Maintainers on the plugin page: https://plugins.jenkins.io/junit/
Do you know anybody who is a maintainer?

jglick:
> your name is the only one listed under Maintainers on the plugin page

Perhaps because I happened to do a release recently, to get in #130 since it was observed in ci.jenkins.io and looked both serious and easily fixed. I have no idea who, if anyone, would claim to be maintaining this plugin for purposes of general feature development or minor fixes.

src/test/java/hudson/tasks/junit/TestResultTest.java
```diff
@@ -153,16 +154,20 @@ public void testTestSuiteDistributedOverMultipleFilesIsCountedAsOne() throws IOE
     /**
      * A common problem is that people parse TEST-*.xml as well as TESTS-TestSuite.xml.
      * See http://jenkins.361315.n4.nabble.com/Problem-with-duplicate-build-execution-td371616.html for discussion.
+     * 
+     * UPDATE: de-duplication causes missing tests, and the original use case hasn't worked in several releases.
+     * Therefore, the de-duplication logic has been removed.
      */
+    @Test
     public void testDuplicatedTestSuiteIsNotCounted() throws IOException, URISyntaxException {
```
timja: shouldn't this be:
```suggestion
    public void testDuplicatedTestSuiteIsCounted() throws IOException, URISyntaxException {
```

or simply removed because it's no longer a valid case?

timja: FYI I removed it

khneal: Yes, I'm okay with that. Thanks!

khneal:
THANK YOU @timja !!!
I've thought about this PR several times per week for months, even if I didn't come back to visit as often as I would have liked. If I ever happen to meet you IRL, then the beer is on me (or pretzel, or whatever other token of appreciation works for you). 😃

timja:
ha no problem, hope it works for you 😄.

PR #133, opened 2020-04-28
[JENKINS-34217] add option to disable historical statistics gathering…

XavierRaynaud:
… for test results

This change adds a new job configuration property that, when enabled, disables
the display of the test trend chart and prevents the plugin from even calculating
how long a test has been failing for, thus saving the work of digging up historical
records.
Note that this commit is squashing the 3 commits initially devloped by Jim Carrothers.
It also fixes h.t.j.JobTestResultDisplayProperty$DescriptorImpl.getDisplayName() return value.

### DIFF
--- a/src/main/java/hudson/tasks/junit/JobTestResultDisplayProperty.java
+++ b/src/main/java/hudson/tasks/junit/JobTestResultDisplayProperty.java
@@ -1,0 +1,66 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 SAP SE
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package hudson.tasks.junit;
+
+import hudson.Extension;
+import hudson.model.Job;
+import hudson.model.JobProperty;
+import hudson.model.JobPropertyDescriptor;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ *
+ */
+public class JobTestResultDisplayProperty extends JobProperty<Job<?, ?>> {
+	private final boolean disableHistoricalResults;
+
+	public JobTestResultDisplayProperty(Boolean disableHistoricalResults) {
+		if ( disableHistoricalResults != null ) {
+			this.disableHistoricalResults = disableHistoricalResults;
+		}
+		else {
+			this.disableHistoricalResults = false;
+		}
+	}
+
+	public boolean getDisableHistoricalResults() {
+		return disableHistoricalResults;
+	}
+
+	@Extension
+	public static class DescriptorImpl extends JobPropertyDescriptor {
+
+		@Override
+		public String getDisplayName() {
+			return "";
+		}
+
+		@Override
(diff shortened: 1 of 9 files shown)


timja:
Can you please take a look at the merge conflicts,

Is this still needed after https://github.com/jenkinsci/junit-plugin/pull/134,

It no longer blocks page render, instead it is loaded async

jglick:
Also #142 should obviate the need for such workarounds.

qlihdunn:
This bug is killing our CI - Interested to help merge.  Can anyone help or get me started - if I can get time allocated?

src/main/java/hudson/tasks/test/TestResultProjectAction.java
```diff
@@ -156,7 +156,7 @@ private TestResultActionIterable createBuildHistory(Run<?, ?> lastCompletedBuild
     @Deprecated
     public void doTrend( StaplerRequest req, StaplerResponse rsp ) throws IOException, ServletException {
         AbstractTestResultAction a = getLastTestResultAction();
-        if(a!=null)
+        if(a!=null && a.shouldCalculatePreviousResults())
```
timja, line 159: this is not used anymore

src/main/resources/hudson/tasks/test/TestResultProjectAction/floatingBox.jelly
```diff
@@ -25,5 +25,31 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:local="local" xmlns:c="/charts">
-  <c:trend-chart it="${from}" title="${%Test Result Trend}" enableLinks="true"/>
+  <j:set var="tr" value="${action.lastTestResultAction}" />
```
timja, line 28: incorrect merge conflict resolution

src/main/java/hudson/tasks/junit/JobTestResultDisplayProperty.java
```diff
@@ -0,0 +1,66 @@
...
+
+		@Override
+		public String getDisplayName() {
+			return "";
+		}
+
+		@Override
+		public JobTestResultDisplayProperty newInstance(StaplerRequest req, JSONObject formData) {
+			if (formData.isNullObject()) return null;
+			return new JobTestResultDisplayProperty(formData.getBoolean("junitsettings-disableHistoricalResults"));
```
timja, line 63: field should be stored on the descriptor I think and then use data binding instead of retrieving from formData

src/main/java/hudson/tasks/junit/JobTestResultDisplayProperty.java
```diff
@@ -0,0 +1,66 @@
...
+import hudson.model.JobProperty;
+import hudson.model.JobPropertyDescriptor;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ *
+ */
+public class JobTestResultDisplayProperty extends JobProperty<Job<?, ?>> {
+	private final boolean disableHistoricalResults;
```
timja, line 37: use spaces not tabs please to be consistent with elsewhere

timja (changes requested):
I've done a sanity review and not tested it locally, it won't work as is.

I would strongly suggest people check out https://github.com/jenkinsci/junit-sql-storage-plugin if this is causing an issue for them. 

@XavierRaynaud / @qlihdunn what is the reason you want this feature? the issue talks about a specific gerrit feature, is your use-case the same?

XavierRaynaud:
Hi @timja,

Many thanks for your comment.
Frankly speaking, I just retrieved Jim Carrothers's patch, and tried to apply it. It seems that it requires a bit more work :)

The reason I want this feature is the following:
(1) I have jenkins jobs with a huge amount of junit test records.
(2) Moreover, some tests in our jobs are lazyly run (they are performed only when corresponding code has been modified).

Because of (1), trend chart, statistics, and browsing test results tooks a huge time.
Because of (2), trend chart, statistics, and browsing in test history is often meaningless.

Therefore, I would like to cut down all history computation for such jobs.

timja:
> (1) I have jenkins jobs with a huge amount of junit test records.

https://github.com/jenkinsci/junit-sql-storage-plugin should help with this.

and it should make it so that 2 is not such a problem

XavierRaynaud:
Ok, thanks for the info.

I had a look on the code, and did not see anything about DELETE statements.
Does it means that junit results stored in database are never deleted, even when run are deleted ?

timja:
> Ok, thanks for the info.
> 
> I had a look on the code, and did not see anything about DELETE statements.
> Does it means that junit results stored in database are never deleted, even when run are deleted ?

not currently deleted as they are also used for historical statements, run auto-rotation doesn't mean that the test results are no longer interesting.

but some form of cleaning should be implemented

XavierRaynaud:
Ok. thanks.

As I said before, I've a huge number of junit tests.
Moreover, stdout/stderr of such tests can be huge too...

Therefore, I can't switch to a database if the policy is to keep results forever.

timja:
> Ok. thanks.
> 
> As I said before, I've a huge number of junit tests.
> Moreover, stdout/stderr of such tests can be huge too...
> 
> Therefore, I can't switch to a database if the policy is to keep results forever.

It now deletes them when the job or build is deleted https://github.com/jenkinsci/junit-sql-storage-plugin/releases/tag/99.v79279ea1504d

XavierRaynaud:
Wow cool ! Thanks a lot. I will test it asap. I think we can close this pull request.

PR #134, opened 2020-06-24
Convert trend to ECharts

timja:
Hi 👋 

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

There's one behaviour change in that it shows passing tests instead of total tests,
thoughts?

This was based off of https://github.com/jenkinsci/warnings-ng-plugin/blob/8ff937365caf3559c19a1e592aa218359a278347/plugin/src/main/java/io/jenkins/plugins/analysis/core/model/JobAction.java#L181

cc @oleg-nenashev @jglick @uhafner @fqueiruga

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.1</version>
+        <version>4.3</version>
         <relativePath />
     </parent>
     <artifactId>junit</artifactId>
@@ -15,7 +15,7 @@
     <properties>
         <revision>1.30</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.241</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
     </properties>
@@ -44,6 +44,11 @@
         </pluginRepository>
     </pluginRepositories>
     <dependencies>
+        <dependency>
+            <groupId>io.jenkins.plugins</groupId>
+            <artifactId>echarts-api</artifactId>
+            <version>4.7.0-4</version>
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -123,6 +123,11 @@
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
+            <dependency>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-annotations</artifactId>
+                <version>4.0.2</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -118,8 +118,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.164.x</artifactId>
-                <version>9</version>
+                <artifactId>bom-2.204.x</artifactId>
+                <version>11</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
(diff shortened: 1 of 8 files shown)


fqueiruga:
@uhafner Does the echarts library brings in bootstrap? I'm a bit concerned to put a full bootstrap page on the main build page. 

Other than that, it looks quite good.

timja:
No it doesn't:

image

fqueiruga:
That's great then!

uhafner:
No, I does not use bootstrap. (It requires jQuery3 plugin up to now).

pom.xml
```diff
@@ -15,7 +15,7 @@
     <properties>
         <revision>1.30</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.204.4</jenkins.version>
```
jglick, line 18: switch `bom` import

jglick:
> it shows passing tests instead of total tests

If I understand the “after” chart it seems like there may be some color distinction between red and purple (~ red + blue)? If so, this is really hard for me to grasp, and I cannot imagine how you would overlay “skipped” on that unless you switched from grey to yellow, at which point you are in a three-dimensional color space… Suggest returning to the original format, which I at least find easier to interpret.

timja:
Thoughts now?

image

skundrik:
If you want to go fancy you can add a filter by legend. Click to toggle on graph.

uhafner:
> If you want to go fancy you can add a filter by legend. Click to toggle on graph.

This is automatically provided by echarts legend.

timja:
> > If you want to go fancy you can add a filter by legend. Click to toggle on graph.
> 
> This is automatically provided by echarts legend.

Didn't notice that before, that's nice!

timja:
@jglick

timja:
> I like this visually. How will this play together with other plugins visualizing test results, like matrix-project?

does it call the junit APIs?

If so I've those APIs to be replaced with the echarts ones

olivergondza:
To the best of my understanding, some at least are referring to generated image: https://github.com/search?q=org%3Ajenkinsci+trendMap&type=Code

Keeping the endpoint in place keeps them from breaking, but the maintainers might not realize this is actually deprecated as they are not calling the Java API. My question was really to see if there is an effort to update all the plugins to avoid breakage and unify the appearance.

timja:
> To the best of my understanding, some at least are referring to generated image: [github.com/search?q=org%3Ajenkinsci+trendMap&type=Code](https://github.com/search?q=org%3Ajenkinsci+trendMap&type=Code)
> 
> Keeping the endpoint in place keeps them from breaking, but the maintainers might not realize this is actually deprecated as they are not calling the Java API. My question was really to see if there is an effort to update all the plugins to avoid breakage and unify the appearance.

To a degree yes, but focusing on the most used ones right now...

If I find time I'll try update one other as an example

uhafner:
Matrix project (and Maven project) do not use a separate UI they just use the charts from the JUnit plugin automatically. 
And the `trendMap` methods are provided by each plugin. I think no other plugin is using these methods of the JUnit plugin. They just use the same copy of the code that produces the maps. These methods should be private within a plugin (they can't since Stapler is not very smart with mapping methods).

olivergondza:
Cool, thanks!

timja:
Matrix project seems to work fine:

image

timja:
All seems fine unless someone can point me to something in particular as I've never used those plugins

timja:
@jglick @oleg-nenashev do you need anything else from me?

Thanks 🙏

olivergondza:
I am fine with the changes the way they are. Unfortunately, there might be no active maintainer to integrate this so you might very well go ahead and integrate this yourself as soon as there is a consensus.

timja:
> I am fine with the changes the way they are. Unfortunately, there might be no active maintainer to integrate this so you might very well go ahead and integrate this yourself as soon as there is a consensus.

Core team doesn't appear to have access, but happy to request permissions, if there's no active maintainer here, (not looking to do much other than fix any fallout if there's any and small patches possibly).

PR #141, opened 2020-08-15
External storage - Part 1

timja:
Subsumes https://github.com/jenkinsci/junit-plugin/pull/110

The trend graph and most of the test report work.
Although trend graph could probably be rewritten to not load runs and instead use a query

History graph needs rewriting but can be done standalone

Package results aren't being loaded correctly in the test report

Blue ocean test reporting needs checking at some point.

I plan to continue it in separate PRs


src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -101,6 +104,46 @@ private static float parseTime(Element testCase) {
...
+
+    /**
+     * Used to create a fake failure, when Hudson fails to load data from XML files.
+     * Public since 1.526.
+     *
+     * @param parent Parent result.
+     * @param testName Test name.
+     * @param errorStackTrace Error stack trace.
+     */
+    public CaseResult(SuiteResult parent, String testName, String errorStackTrace) {
```
timja, line 116: these are just moved up so that all the constructors are together

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -173,6 +155,7 @@ public void twoSteps() throws Exception {
         FilePath secondTestFile = ws.child("second-result.xml");
         secondTestFile.copyFrom(TestResultTest.class.getResource("junit-report-2874.xml"));
 
+        rule.pause();
```
timja, line 158:
```suggestion
```

src/test/java/hudson/tasks/junit/storage/TestResultStorageTest.java
```diff
@@ -0,0 +1,649 @@
...
+            assertEquals(1, packageSkippedTests.size());
+            assertEquals("other.Klazz", packageSkippedTests.get(0).getClassName());
+            assertEquals("Not actually run.", packageSkippedTests.get(0).getSkippedMessage());
+
+            PackageResult root = a.getResult().byPackage("(root)");
+            List<CaseResult> rootPassedTests = root.getPassedTests();
+            assertEquals(1, rootPassedTests.size());
+            assertEquals("Klazz", rootPassedTests.get(0).getClassName());
+
+            r.pause();
```
timja, line 193:
```suggestion
```

src/test/java/hudson/tasks/junit/storage/TestResultStorageTest.java
```diff
@@ -0,0 +1,647 @@
...
+        }
+    }
+
+    @TestExtension public static class Impl implements TestResultStorage {
+
+        static final String CASE_RESULTS_TABLE = "caseResults";
+
+        static boolean queriesPermitted;
+
+        private final ConnectionSupplier connectionSupplier = new LocalConnectionSupplier();
```
jglick, line 208: I see that this impl has been copied to https://github.com/timja/jenkins-junit-postgresql-plugin/blob/c83e4553ae625e5976f5addaf617402068601266/src/main/java/io/jenkins/plugins/junit/postgresql/PostgresTestResultStorage.java#L39; does it make sense to define a SQL-based implementation in `junit-plugin/src/main/`? Or to define a `junit-sql` plugin with that implementation and add to `test` scope?

jglick:
> trend graph could probably be rewritten to not load runs and instead use a query

This is a critical piece, I think, as it is an expensive operation done by default on the job index page, and currently limited to (IIRC) 20 builds just to avoid trashing disk.

timja:
@jglick: https://github.com/jenkinsci/junit-plugin/pull/143, probably needs something for limiting build number though

src/test/java/hudson/tasks/junit/storage/TestResultStorageTest.java
timja: Possibly tbh I’m not using any Postgres specific features so could just call it junit-sql plugin not Postgres.

It’s very convenient to be able to develop the pluggable storage from inside the junit plugin, possibly once it matures it could be removed from here and then the external reference implementation added as test scope

timja:
I'm going to proceed with merging to make iterating easier in smaller PRs, but feel free:

* to review after the fact
* to comment on https://github.com/jenkinsci/junit-plugin/issues/142
* to raise new issues for discussion

Issue #142, opened 2020-08-15
Pluggable storage


timja:
In common CI/CD use-cases a lot of the space is being consumed by test reports. This data is stored within JENKINS_HOME, and the current storage format requires huge overheads when retrieving statistics and, especially, trends. In order to display trends, each report has to be loaded and then processed in-memory.

The main purpose of externalising Test Results is to optimize Jenkins logic by querying the desired data from specialized external storages, e.g. from Document-based databases like Elasticsearch. According to the current plan, JUnit Plugin will be extended in order to support such external storage in its APIs being widely used by test reporting plugins.

## Status:

Foundation work started

Prototype API: https://github.com/jenkinsci/junit-plugin/pull/110

Continued in: https://github.com/jenkinsci/junit-plugin/pull/141

## Todo:

- [ ] New trend chart implementation that doesn't rely on loading Runs
- [ ] Rewrite history chart, probably port it to echarts
- [ ] Fix package result loading showing no results when viewing by package
- [ ] Plugin implementing API - most likely a PostgreSQL plugin
- [ ] Benchmarks - show performance improvements for large instances
- [ ] Present at Cloud native Sig
- [ ] Announcement blog on Jenkins.io

## Test strategy:

The junit plugin will provide a sample implementation using h2 database that can be used to verify pluggable storage functionality.

The Postgres implementation will be able to use something like test containers and replicate a similar test inside it's own plugin.

Performance tests will be written using the Microbenchmark harness, see https://www.jenkins.io/blog/2019/06/21/performance-testing-jenkins/

## Graduation criteria

This API will be released in Beta initially and will likely change as it's developed.

* At least one production ready pluggable storage implementation
* Benchmarks showing improvements
* Blog on Jenkins.io

oleg-nenashev:
FTR https://www.jenkins.io/sigs/cloud-native/pluggable-storage/

@timja Thanks for re-starting it! Do you have a particular database in mind for the reference implementation?

timja:
Postgres: https://github.com/timja/jenkins-junit-postgresql-plugin

oleg-nenashev:
I wonder whether we could unify the Data serialization logic with https://github.com/jenkinsci/postgresql-fingerprint-storage-plugin . CC @stellargo

jglick:
> Test result storage also has extensible data structure

Yes but a very different structure. I do not think it would be productive to mix those.

jglick:
[...]

PR #143, opened 2020-08-15
Pluggable storage trend chart

timja:
Requires https://github.com/jenkinsci/junit-plugin/pull/141

Part of https://github.com/jenkinsci/junit-plugin/issues/142


jglick:
From https://github.com/jenkinsci/junit-plugin/pull/141#issuecomment-675030721 not sure I understand

> needs something for limiting build number

For background, originally the trend displayed a graph of all builds, no problem. Then in Jenkins 1.484 (IIRC) lazy loading was introduced. I soon found that the trend graph forced every build to be loaded, so I capped the number of builds considered. If we have, say, SQL storage of results then there is no reason not to go back to showing the complete graph.

timja:
Ok sure, then no need to do that then.

This works and is super quick (as you would expect from a sql based implementation)

timja:
I'm going to proceed with merging to make iterating easier in smaller PRs, this PR is likely going to be needed in a couple of others.

 but feel free:

* to review after the fact
* to comment on https://github.com/jenkinsci/junit-plugin/issues/142
* to raise new issues for discussion

PR #146, opened 2020-08-19
Mark types in hudson.tasks.junit.storage package as beta

jglick:
Not all the hard problems in #142 have been worked on (I suppose), so we should not advertise these interfaces to rando plugin developers without a warning.


timja (approved):
ty, was on my plan before release 

❤️

PR #135, opened 2020-07-16
JENKINS-63113 use (?) as test name when empty in original report file

mguillem:
Fixes https://issues.jenkins-ci.org/browse/JENKINS-63113

The UI for a testcase with an empty name looks then like this:
fix overview

fix details

PR #152, opened 2020-09-03
Avoid NPE from CaseResult.getName()

jglick:
See https://github.com/jenkinsci/junit-plugin/pull/135#issuecomment-686469488. CC @mguillem

### DIFF
--- a/src/test/java/hudson/tasks/junit/CaseResultTest.java
+++ b/src/test/java/hudson/tasks/junit/CaseResultTest.java
@@ -271,6 +271,14 @@ public class CaseResultTest {
         return rule.assertBuildStatus(Result.UNSTABLE, p.scheduleBuild2(0).get());
     }
 
+    @Test
+    public void emptyName() throws Exception {
+        FreeStyleProject p = rule.createFreeStyleProject();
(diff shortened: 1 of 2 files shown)


timja (approved):

jglick:
Yes I would like to do that. Looks like I have release perms. Thanks!

jglick:
https://github.com/jenkinsci/junit-plugin/releases/tag/junit-1.33

src/test/java/hudson/tasks/junit/CaseResultTest.java
```diff
@@ -271,6 +271,14 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
         return rule.assertBuildStatus(Result.UNSTABLE, p.scheduleBuild2(0).get());
     }
 
+    @Test
+    public void emptyName() throws Exception {
```
mguillem, line 275: `missingName` instead of `emptyName`?

jglick: Yeah I suppose so.

PR #157, opened 2020-09-14
History chart port to echarts

timja:


src/main/resources/hudson/tasks/junit/History/index.jelly
```diff
@@ -24,41 +24,40 @@ THE SOFTWARE.
...
+								<span class="sr-only">Previous</span>
+							</a>
+							<a class="carousel-control-next" data-target="#trend-carousel" role="button" data-slide="next">
+								<span class="carousel-control-next-icon" aria-hidden="true"/>
+								<span class="sr-only">Next</span>
+							</a>
+						</div>
+					</bs:card>
+					
+					<c:trend-chart it="${it}" title="${%title(it.testObject.displayName)}" enableLinks="true"/>
```
uhafner, line 60: Is this another trend chart? That one loads the JS libs as well. Maybe this causes an ordering problem?

uhafner: This trend chart actually is intended to be used on the build page only...

timja: ah sure no I just left that in it's not needed

timja: didn't help

uhafner: Hmm... it looks good, but this JS/Jelly errors are so hard to spot...

src/main/resources/hudson/tasks/junit/History/index.jelly
```diff
@@ -75,5 +73,7 @@ THE SOFTWARE.
 	      </j:if>	       
       	</div>
     </l:main-panel>
-	</l:layout>
+	</bs:layout>
```
uhafner, line 76: I have the scripts within layout

timja: I'll push that but if I have it there i get:

image

Which I think it caused by the JavaScript running before the DOM is present:

```
    const chartPlaceHolder = document.getElementById(chartDivId);
    const chart = echarts.init(chartPlaceHolder);
```

`chartPlaceHolder` is undefined from zoomable-trend-chart.js

uhafner: Does the card show up? It is placed in an `if`?

timja: card shows up yes.

image

timja: hmm it seems to run multiple times and sometimes it's there and sometimes it's now =/

src/main/webapp/history/history.js
```diff
@@ -0,0 +1,34 @@
...
+        
+        /**
+         * Creates a build trend chart that shows the number of issues for a couple of builds.
+         * Requires that a DOM <div> element exists with the ID '#severities-trend-chart'.
+         */
+        view.getTestResultTrend(function (lineModel) {
+            echartsJenkinsApi.renderZoomableTrendChart('test-result-trend-chart', lineModel.responseJSON, redrawTrendCharts);
+        });
+    }
+})(jQuery3);
```
uhafner, line 34: You need to redraw the chart if the carousel changes the active element. I copied my code here (it also persists the carousel position).

```
/* global jQuery3, view, echartsJenkinsApi */
(function ($) {
    redrawTrendCharts();
    storeAndRestoreCarousel('trend-carousel');

    /**
     * Activate tooltips.
     */
    $(function () {
        $('[data-toggle="tooltip"]').tooltip();
    });

    /**
     * Redraws the trend charts. Reads the last selected X-Axis type from the browser local storage and
     * redraws the trend charts.
     */
    function redrawTrendCharts() {

        /**
         * Creates a build trend chart that shows the number of issues per tool.
         * Requires that a DOM <div> element exists with the ID '#tools-trend-chart'.
         */
        view.getTestDurationTrend(function (lineModel) {
            echartsJenkinsApi.renderZoomableTrendChart('test-duration-trend-chart', lineModel.responseJSON, redrawTrendCharts);
        });
        
        /**
         * Creates a build trend chart that shows the number of issues for a couple of builds.
         * Requires that a DOM <div> element exists with the ID '#severities-trend-chart'.
         */
        view.getTestResultTrend(function (lineModel) {
            echartsJenkinsApi.renderZoomableTrendChart('test-result-trend-chart', lineModel.responseJSON, redrawTrendCharts);
        });
    }

    /**
     * Store and restore the selected carousel image in browser's local storage.
     * Additionally, the trend chart is redrawn.
     *
     * @param {String} carouselId - ID of the carousel
     */
    function storeAndRestoreCarousel (carouselId) {
        const carousel = $('#' + carouselId);
        carousel.on('slid.bs.carousel', function (e) {
            localStorage.setItem(carouselId, e.to);
            const chart = $(e.relatedTarget).find('>:first-child')[0].echart;
            if (chart) {
                chart.resize();
            }
        });
        const activeCarousel = localStorage.getItem(carouselId);
        if (activeCarousel) {
            carousel.carousel(parseInt(activeCarousel));
        }
    }
})(jQuery3);
```

timja: ah thanks

timja:
@uhafner any idea on:

image

I think it's caused by the page getting loaded 4 times and the first time the div's aren't there.

I can't find any reason for the page getting called that many times.

It seems fine from the user experience currently except for the extra AJAX requests and the browser console errors

PR #161, opened 2020-09-17
Fix missing sidebar with pluggable storage on test report

timja:


src/test/java/io/jenkins/plugins/junit/storage/TestResultStorageJunitTest.java
```diff
@@ -220,6 +220,8 @@
 
             final List<TestDurationResultSummary> testDurationResultSummary = pluggableStorage.getTestDurationResultSummary();
             assertThat(testDurationResultSummary.get(0).getDuration(), is(200));
+            
+            r.pause();
```
timja, line 224:
```suggestion
```

PR #170, opened 2020-09-24
Pluggable storage: Add stdout, stderr, stacktrace, fix caseresult loading

timja:
.

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -134,14 +134,24 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     }
 
     @Restricted(Beta.class)
-    public CaseResult(SuiteResult parent, String className, String testName, String errorDetails, String skippedMessage, float duration) {
+    public CaseResult(
+            SuiteResult parent, 
+            String className, 
+            String testName, 
+            String errorDetails, 
+            String skippedMessage, 
+            float duration,
+            String stdout,
+            String stderr,
+            String stacktrace
+    ) {
         this.className = className;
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -134,14 +134,24 @@ public CaseResult(SuiteResult parent, String testName, String errorStackTrace, S
     }
 
     @Restricted(Beta.class)
-    public CaseResult(SuiteResult parent, String className, String testName, String errorDetails, String skippedMessage, float duration) {
+    public CaseResult(
```
timja, line 137: ignoring compat break as it's beta and only user is sql plugin which will be released at the same time

src/main/java/hudson/tasks/junit/PackageResult.java
```diff
@@ -148,15 +145,7 @@ public int getSkipCount() {
...
-        JunitTestResultStorage storage = JunitTestResultStorage.find();
-        ClassResult result;
-        if (!(storage instanceof FileJunitTestResultStorage)) {
-            Run<?, ?> run = Stapler.getCurrentRequest().findAncestorObject(Run.class);
-            TestResultImpl pluggableStorage = storage.load(run.getParent().getFullName(), run.getNumber());
-            result = pluggableStorage.getClassResult(name);
-        } else {
-            result = getClassResult(name);
-        }
+        ClassResult result = getClassResult(name);
```
timja, line 148: get by package result also loads the full class result, this code before was ignoring the package and only worked in tests because we had a Klazz and a other.Klazz. and so the wrong Klazz was being picked.

Now we load from the correct package

PR #173, opened 2020-10-06
Handle stacktrace when no builds in trend chart

timja:
Fixes https://github.com/jenkinsci/junit-plugin/issues/150

### DIFF
--- a/src/main/java/hudson/tasks/test/TestResultProjectAction.java
+++ b/src/main/java/hudson/tasks/test/TestResultProjectAction.java
@@ -26,6 +26,7 @@ package hudson.tasks.test;
 import edu.hm.hafner.echarts.ChartModelConfiguration;
 import edu.hm.hafner.echarts.JacksonFacade;
 import edu.hm.hafner.echarts.LinesChartModel;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/tasks/test/TestResultProjectAction.java
```diff
@@ -118,18 +119,28 @@ protected LinesChartModel createChartModel() {
             TestResultImpl pluggableStorage = storage.load(lastCompletedBuild.getParent().getFullName(), lastCompletedBuild.getNumber());
             return new TestResultTrendChart().create(pluggableStorage.getTrendTestResultSummary());
         }
-        
-        return new TestResultTrendChart().create(createBuildHistory(lastCompletedBuild), new ChartModelConfiguration());
+
+        TestResultActionIterable buildHistory = createBuildHistory(lastCompletedBuild);
+        if (buildHistory == null) {
+            return new LinesChartModel();
```
oleg-nenashev, line 125: Might be better to have a new implementation with "No Build History" text later.

oleg-nenashev (approved):
Looks about right. Later it might require different implementation for Multi-Branch jobs, when the historical data can be inherited from the main branch. But we are not there yet

Issue #160, opened 2020-09-17
Integrate with checks-api plugin


timja:
This will allow publishing test results to GitHub checks

See https://github.com/jenkinsci/checks-api-plugin/blob/master/docs/consumers-guide.md

and https://github.com/jenkinsci/warnings-ng-plugin/pull/550

This should be updated in the JunitStep https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java

timja:
cc @XiongKezhi in case you want to add anything

XiongKezhi:
LGTM now. I can help with it after this hiring season if nobody is still interested in it by then.

KalleOlaviNiemitalo:
FYI, I filed [JENKINS-64714](https://issues.jenkins.io/browse/JENKINS-64714) requesting a similar feature in the [xUnit plugin](https://plugins.jenkins.io/xunit/).

PR #180, opened 2020-10-10
Checks API integration

timja:
Fixes https://github.com/jenkinsci/junit-plugin/issues/160

TODO:

* [x] Add to jelly files
* [x] Tests
* [x] Documentation
* [x] Test freestyle builds
* [x] Use display url API for test report

Example reports can be found in:

https://github.com/timja-org/junit-attachments-test/pulls

A good one is:
https://github.com/timja-org/junit-attachments-test/pull/4/checks?check_run_id=1235203424

and:
https://github.com/timja-org/junit-attachments-test/pull/5


src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
```diff
@@ -0,0 +1,143 @@
...
+        publisher.publish(extractChecksDetails());
+    }
+
+    @VisibleForTesting
+    ChecksDetails extractChecksDetails() {
+        String text = extractChecksText();
+        System.out.println(text);
+        ChecksOutput output = new ChecksOutput.ChecksOutputBuilder()
+                .withTitle(extractChecksTitle())
+                .withSummary("<sub>Send us [feedback](https://github.com/jenkinsci/junit-plugin/issues)")
```
timja, line 49: I was wondering about maybe having a Feedback <- survey, and a bug report link, thoughts?

oleg-nenashev: Whatever works best for you. Just GitHub issues might be enough for developers, but a survey is better for common users

oleg-nenashev (approved):
Looks very promising 👍 . I have not tested the report generation, but I do not see anything wrong in the code. Looking forward to try it out

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
```diff
@@ -0,0 +1,143 @@
...
+            List<CaseResult> failedTests = action.getResult().getFailedTests();
+            
+            if (failedTests.size() > 10) {
+                failedTests = failedTests.subList(0, 9);
+            }
+            
+            failedTests.forEach(failedTest -> mapFailedTestToTestReport(builder, failedTest));
+            if (summary.getFailCount() > 10) {
+                builder.append("\n")
+                        .append(summary.getFailCount() - 10).append(" more test results are not shown here, view them on Jenkins");
```
uhafner, line 74: It would be nice to have Jenkins as hyperlink to the results

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
```diff
@@ -0,0 +1,143 @@
...
+
+        return builder.toString();
+    }
+
+    private void mapFailedTestToTestReport(StringBuilder builder, CaseResult failedTest) {
+        builder.append("## <code>").append(failedTest.getTransformedFullDisplayName().trim()).append("</code>")
+                .append("\n");
+
+        if (StringUtils.isNotBlank(failedTest.getErrorDetails())) {
+            builder.append("<pre><code>").append(failedTest.getErrorDetails().trim()).append("</code></pre>")
```
uhafner, line 86: Is this valid markdown? I would expect 3 ticks for code blocks.

uhafner (approved):
Thanks for adding that!

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
timja: it works, example reports in the description, I'll check if I can simplify it though.

thanks for the suggestion!

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
timja: true, although one line down in the report is the github link to the results

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
```diff
@@ -0,0 +1,148 @@
...
+            if (failedTests.size() > 10) {
+                failedTests = failedTests.subList(0, 9);
+            }
+            
+            failedTests.forEach(failedTest -> mapFailedTestToTestReport(builder, failedTest));
+            if (summary.getFailCount() > 10) {
+                builder.append("\n")
+                        .append(summary.getFailCount() - 10)
+                        .append(" more test results are not shown here, view them on [Jenkins](")
+                        .append(testsURL).append(")");
```
XiongKezhi, line 78: Does this `testsURL` links to the JUnit summary page? If so, I think we should use different links for here and the `detailsURL`, maybe just the build page links for the `detailsURL`?

XiongKezhi (approved):

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
timja: It links to the same place, I don’t think there’s any need to link differently the only place one would go it the test report from this check

vladaurosh:
Hello.
Can we add option to remove messages "[Checks API] No suitable checks publisher found." ?

timja:
the option is there and documented in https://github.com/jenkinsci/junit-plugin#test-result-checks-for-github-projects
`skipPublishingChecks: true`

XiongKezhi:
I guess he wants to remove the status checks logs at all @timja 

@vladaurosh And actually, that's something we are doing now: https://github.com/jenkinsci/checks-api-plugin/pull/31

timja:
skipping publishing checks will do that, (although it would be nice to have a global switch or at least at the job level to opt out of the log)

vladaurosh:
Thanks @timja and @XiongKezhi .
I'll wait for checks-api-plugin#31 to be done. :)

vasilvestre:
I don't get how I should configure the publisher so my results are in github. Documentation doesn't really helps me here, any help is welcome

PR #189, opened 2020-10-21
Allow setting custom checks name

mrginglymus:
Adds the ability to set a custom name for the Checks API for each junit invocation.

I could see a number of ways to skin this cat, so I've picked this one on a fairly arbitrary basis.

This is also my first contribution to a jenkins plugin so please forgive anything I've missed!

### DIFF
--- a/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
+++ b/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
@@ -1,0 +1,4 @@
+<div>
+    If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
+    SCM hosting platforms. If not, a default of "Test" will be used.
+</div>
--- a/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
+++ b/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
@@ -1,4 +1,4 @@
 <div>
     If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
-    SCM hosting platforms. If not, a default of "Test" will be used.
+    SCM hosting platforms. If not, a default of "Tests" will be used.
 </div>
--- a/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
+++ b/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
@@ -1,0 +1,4 @@
+<div>
+    If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
+    SCM hosting platforms. If not, a default of "Test" will be used.
+</div>
--- a/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
+++ b/src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
@@ -1,4 +1,4 @@
 <div>
     If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
-    SCM hosting platforms. If not, a default of "Test" will be used.
+    SCM hosting platforms. If not, a default of "Tests" will be used.
 </div>
--- a/src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
+++ b/src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
@@ -50,6 +50,7 @@ public class JUnitChecksPublisherTest {
         ChecksDetails checksDetails = publisher.extractChecksDetails();
 
         assertThat(checksDetails.getConclusion(), is(ChecksConclusion.SUCCESS));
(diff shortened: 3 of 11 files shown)


oleg-nenashev (approved):
This is quite straightforward, and it should do the job. Would be also great to add this feature to the checks documentation in https://github.com/jenkinsci/junit-plugin#test-result-checks-for-github-projects

P.S: maybe we should consider changing the default from "Tests" to "Unit Tests"

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -114,4 +115,32 @@ public void extractChecksDetailsFailingMultipleTests() throws Exception {
...
         assertThat(output.getTitle().get(), is("failed: 3, passed: 5"));
     }
+
+    @Test
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
+    public void setCustomCheckName() throws Exception {
+        WorkflowJob j = rule.jenkins.createProject(WorkflowJob.class, "singleStep");
+        j.setDefinition(new CpsFlowDefinition("stage('first') {\n" +
+              "  node {\n" +
+              "    def results = junit(testResults: '*.xml', checksName: 'Custom Checks Name')\n" + // node id 7
```
timja, line 125: that comment is not needed here, (or above if you want to remove it)

```suggestion
              "    def results = junit(testResults: '*.xml', checksName: 'Custom Checks Name')\n" +
```

timja (commented):
Missing changes in https://github.com/jenkinsci/junit-plugin/blob/74da9ea7d4855eba18a25e0ecd35e4827d763146/src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly

that will allow pipeline syntax generator to work and freestyle builds

timja:
> This is quite straightforward, and it should do the job. Would be also great to add this feature to the checks documentation in [jenkinsci/junit-plugin#test-result-checks-for-github-projects](https://github.com/jenkinsci/junit-plugin#test-result-checks-for-github-projects)

👍 

> P.S: maybe we should consider changing the default from "Tests" to "Unit Tests"

Well, a lot of people will be using it for others, so probably not worth changing it.

src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
+    SCM hosting platforms. If not, a default of "Test" will be used.
```
timja, line 3:
```suggestion
    SCM hosting platforms. If not, a default of "Tests" will be used.
```

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
+    SCM hosting platforms. If not, a default of "Test" will be used.
```
timja, line 3:
```suggestion
    SCM hosting platforms. If not, a default of "Tests" will be used.
```

timja (approved):
minor nit, LGTM pending a quick manual test for sanity

timja:
I've tested it out, some minor cleanup to make snippet generator a bit nicer.

I'll merge and release after CI is green

Thanks!

timja:
Could you file an issue for that along with a sample of the mechanism you describe? 

And what that would look like to users...

I don’t have time to work on that atm but fine to review and merge if someone contributes a replacement

mrginglymus:
While I don't disagree that there may be a better way of detecting a default value that isn't `Tests`, I would be incredibly disappointed to see the ability for a user to set their own checks name at the point of junit invocation removed.

Providing it via the program that produces the xml to be consumed would require both substantial changes to our Jenkinsfiles and junit producing programs, compared to the one line we are able to achieve the behaviour with currently.

We already have an issue with the warningsng plugin where the name of the report on jenkins has to be the exact same as the check reported to github.

For context, our use case here is that we have a number of parent jobs that may run on a single PR, each with a large number of inidividual checks - e.g. our smoke run has > 10 checks reported, our long run about 5. They are a mix of test results, warnings, and coverage, plus some custom checks. They are also duplicated _between_ parent jobs, for instance our build stage runs on both the long run and the deploy run.

Therefore, to help keep some kind of order to the large number of checks, we namespace them by the parent job, so we end up with:

```
Deploy  # github status, links to jenkins
Deploy / Build  # github check, links to github
Deploy / Deploy
Smoke
Smoke / Linting
Smoke / Unit Tests
Long
Long / Build
Long / Long Tests
```

etc.

The easiest way to achieve this, by far, is for every plugin that publishes a check to allow the user to provide the name of that check.

I'm also hoping that at some point the checks API will support _updating_ checks, so that we can have a pattern that looks like:

```
withCheck('My Check Name') { check ->
    sh 'test --output output.xml'
    junit testResults: 'output.xml', check: check
}
```

ie, a wrapper that creates a check in progress, catches and reports exceptions.

PR #190, opened 2020-10-21
Add ability to ignore current stage name when computing test names 

mrginglymus:
Opening this PR as a way to demonstrate the problem I'm trying to fix:

I currently have a pipeline that looks something like this:
```
-- unit tests -- / -- integration chunking -- \ -- chunk 1 -- / -- integration test reporting --
                                              |               |
                                              \ -- chunk 2 -- /
                                              |               |
                                              \ -- chunk 3 -- /
```

That is, we run some fast unit tests, and then split up our slow integration tests to be run in chunks on multiple agents, in parallel. This chunking is optimised for evenly sized bins, so which tests are in which chunks will change on each run. At the end of each chunk we stash the junit xml file, and in a final stage once all chunks have run we unstash them all and invoke junit.

The reason we have to do it this convoluted way is that if we were to invoke junit in the chunks, we would end up with the chunk name in the test report, which would make tracking test stability over time effectively impossible.

While this workaround is serving us reasonably well, it is quite annoying when a single test fails and you have to do the legwork to find out which of the chunks the test failed in, so you can download the relevant artifacts and do some sniffing.

This PR, which I do not claim to be the right solution to this problem, merely _a_ solution, helps us with this by adding an option to ignore the innermost stage name when calculating enclosing blocks, effectively re-scoping the junit invocation to the stage containing the parallel invocations.

Could I please have some thoughts on whether this is a completely ridiculous idea and I should continue to just use our workable workaround, whether there's a way of doing this without code change, or any other ideas for how to skin this particular cat.

### DIFF
--- a/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
+++ b/src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStep.java
@@ -53,6 +53,7 @@ public class JUnitResultsStep extends Step implements JUnitTask {
      */
     private boolean allowEmptyResults;
     private boolean skipPublishingChecks;
+    private boolean ignoreCurrentStage;
 
     @DataBoundConstructor
     public JUnitResultsStep(String testResults) {
@@ -129,6 +130,15 @@ public class JUnitResultsStep extends Step implements JUnitTask {
     @DataBoundSetter
     public void setSkipPublishingChecks(boolean skipPublishingChecks) {
         this.skipPublishingChecks = skipPublishingChecks;
+    }
+
+    public boolean isIgnoreCurrentStage() {
+        return ignoreCurrentStage;
+    }
+
+    @DataBoundSetter
+    public void setIgnoreCurrentStage(boolean ignoreCurrentStage) {
+        this.ignoreCurrentStage = ignoreCurrentStage;
     }
 
     @DataBoundSetter public final void setAllowEmptyResults(boolean allowEmptyResults) {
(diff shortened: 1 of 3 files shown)


timja:
Not claiming to fully understand,

but thinking out loud, why does it matter which chunk a test failed in?

all test output, stdin, stdout, stderr will be correctly linked to a test result

artifacts wise you have at least two options:

1. only archive / output them on failed tests
2. use the https://plugins.jenkins.io/junit-attachments/ plugin to link artifacts to a test result.

mrginglymus:
Thanks for the feedback!

> but thinking out loud, why does it matter which chunk a test failed in?

Currently our test reports pop out on jenkins to e.g.

```
integration tests / path.to.my.test_function
```

If we ran junit on each chunk, rather than at the end, this would be

```
integration tests / chunk n / path.to.my.test_function
```

where `n` would change each run, which means that we wouldn't be able to track the history of that test as it would come and go depending on how the chunking ran.

> only archive / output them on failed tests

Ah, we slightly shoot ourselves in the foot a bit there. Because we run a number of different test runners, each with their own unique idea of what they should do with exit codes, we force _all_ of them to return an exit code of 0 if they successfully executed tests regardless of success/failure, and use junit to work out which ones actually failed or not (this so we can correctly explode and stop the run if test _execution_ fails). So we don't know if we need to keep the artifacts until it's too late. Perhaps we should reconsider that approach, but even then we'd still hit the above problem with the test naming.

> use the https://plugins.jenkins.io/junit-attachments/ plugin to link artifacts to a test result.

that looks interesting! Unfortunately, especially for integration testing (but even for unit testing) we have a lot of shared artifacts between tests (mostly logging) which we can't split by test; the only way to use this plugin would be to duplicate these artifacts for each failed test as far as I can see.

timja:
I've never used this but maybe it could help? https://github.com/jenkinsci/parallel-test-executor-plugin

Not really sure if your change is the right approach, it seems quite hacky.

Issue #192, opened 2020-10-23
Checks API limit


timja:
need to trim the characters we send to checks API (65535) is the limit:

```
[2020-10-23T15:02:30.961Z] [GitHub Checks] Failed Publishing GitHub checks: org.kohsuke.github.GHFileNotFoundException: https://api.github.com/repos/jenkinsci/jenkins/check-runs {"message":"Invalid request.\n\nOnly 65535 characters are allowed; 569338 were supplied.","documentation_url":"https://docs.github.com/rest/reference/checks#create-a-check-run"}
```

Needs to be shortened here ish https://github.com/jenkinsci/junit-plugin/blob/master/src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java#L76

timja:
fixed by https://github.com/jenkinsci/junit-plugin/pull/193

KalleOlaviNiemitalo:
The 65535-character limit seems to be specific to GitHub.

* GitLab: I did not find a checks API as such. [Post the build status to a commit](https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit) supports a "short description". [Create new merge request note](https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note) has a limit of 1,000,000 characters. [Unit test reports](https://docs.gitlab.com/ee/ci/unit_test_reports.html) reads JUnit reports from build artifacts.
* Bitbucket Server: The details string of a Code Insights report is at most 2000 characters; likewise the message string of each annotation in the report. [REST Resources Provided By: Bitbucket Server - Code Insights](https://docs.atlassian.com/bitbucket-server/rest/7.7.0/bitbucket-code-insights-rest.html)
[...]

PR #193, opened 2020-10-24
Ensure message to check api is under 64kb

babusekaran:
fix for https://github.com/jenkinsci/junit-plugin/issues/192

limit appending to message when it exceeds the message size currently supported by [check API](https://developer.github.com/v3/checks/runs/#annotations-object)


babusekaran:
@timja I have made changes to limit message also based on size, i feel the previous implementation of limiting 10 tests can be removed as now it can create message content based on maximum size defined. WDYT?

timja:
> @timja I have made changes to limit message also based on size, i feel the previous implementation of limiting 10 tests can be removed as now it can create message content based on maximum size defined. WDYT?

yes absolutely that can go now. I don't think we need to log the remaining tests either that was only there because 10, so you can probably convert to a regular for loop, i.e. `for (CaseResult failedTest : failedTests) {`

src/main/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisher.java
```diff
@@ -1,142 +1,145 @@
...
+    }
+
+    private String extractChecksText(String testsURL) {
+        StringBuilder builder = new StringBuilder();
+        if (summary.getFailCount() > 0) {
+            List<CaseResult> failedTests = action.getResult().getFailedTests();
+
+            for (CaseResult failedTest: failedTests) {
+                String testReport = mapFailedTestToTestReport(failedTest);
+                int messageSize = testReport.getBytes(StandardCharsets.UTF_8).length + builder.toString().getBytes(StandardCharsets.UTF_8).length;
```
timja, line 65: is there any reason you're calling get bytes and not just `length()`

babusekaran: Ideally `length()` should work perfectly fine, was thinking if text contains other more Unicode values length may not be equal to byte size

timja: i think length should be fine here. we're also substracting 1000 so should help any edge cases.

babusekaran: makes sense, changed to `length()` 👍

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
timja, line 1: github was showing the whole file as a diff because it got converted to windows file endings, this should prevent that in future and i converted the line endings back

timja (approved):
code looks good, I just need to test it against github to make sure it behaves properly.

timja:
Tested here: https://github.com/timja-org/junit-attachments-test/pull/6/checks?check_run_id=1304317402

PR #199, opened 2020-10-29
Publish only results included in current invocation to checks api - fixes #198

### LINKED ISSUE #198: Mismatch between test summary and tests reported
Having used the ability to have custom checks name in anger, I spotted what I thought was a missing feature, but actually turns out to be appear to be a bug, possibly on this line:
https://github.com/jenkinsci/junit-plugin/blob/d3fd50c99b510a36e31f173c3d497190038a0419/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java#L287

as the `action` here is a merge of all `junit` invocations in a run, but the `summary` is only the count of passes/fails in the _current_ invocation of `junit`.

For example:

```
stage('first stage') {
    junit 'test1.xml' // contains 1 failure, test_foo
}
stage('second stage') {
    junit 'test2.xml' // contains 2 failures, test_bar and test_baz
}
```

Will result in a check being published that will claim that there are only two failures, but list all three of `test_foo`, `test_bar` and `test_baz`.

The behaviour that I would find most useful is to only report in the checks the failures in that current invocation; the fix for this looks simple enough as far as I can tell, and I'll try to put together a PR if that's the option that's desirable.

It's also entirely possible I've not understood the code and not spotted the actual cause, but the effect definitely appears to be there: 

image

mrginglymus:
This, I believe, should fix the problem described in #198. However, I'm at a bit of a loss as to how best to test it, short of mocking a checks publisher and inspecting calls to that.

fixes #198

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -186,7 +186,6 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
                 action.mergeResult(result, listener);
             }
             action.setHealthScaleFactor(task.getHealthScaleFactor()); // overwrites previous value if appending
-            action.setChecksName(task.getChecksName());
             if (result.isEmpty()) {
                 if (build.getResult() == Result.FAILURE) {
                     // most likely a build failed before it gets to the test phase.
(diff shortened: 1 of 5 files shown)


timja:
> This, I believe, should fix the problem described in #198. However, I'm at a bit of a loss as to how best to test it, short of mocking a checks publisher and inspecting calls to that.

I think just manually testing this bit is fine. otherwise yeah verifying the second call only contains the second results and not the combination of both would be the first to go.

mrginglymus:
@timja apologies for the delay here. I've finally had a chance to test this on our jenkins and it does the job 👍

mrginglymus:
Thanks for merging this @timja; do you have a rough idea of when you might next cut a release?

timja:
> Thanks for merging this @timja; do you have a rough idea of when you might next cut a release?

thanks for the nudge, I've triggered it. Should be available in ~15 mins or so

mrginglymus:
Thanks!

Issue #210, opened 2020-11-16
Remove/deprecate checksName parameter to junit step


jglick:
`checksName` was introduced in #189 as an attempt to address #183, namely

> we have multiple invocations of junit (across multiple different pipelines jobs) per PR.
> This means that currently, only the last reported test appears.

It suffices to use `SuiteResult.getEnclosingBlockNames`, which will be set automatically according to `stage` or `parallel` branch as of #76, without modifications to the `Jenkinsfile`.

mrginglymus:
Posted on the PR initially, reposting here:

---

While I don't disagree that there may be a better way of detecting a default value that isn't `Tests`, I would be incredibly disappointed to see the ability for a user to set their own checks name at the point of junit invocation removed.

Providing it via the program that produces the xml to be consumed would require both substantial changes to our Jenkinsfiles and junit producing programs, compared to the one line we are able to achieve the behaviour with currently.

We already have an issue with the warningsng plugin where the name of the report on jenkins has to be the exact same as the check reported to github.

For context, our use case here is that we have a number of parent jobs that may run on a single PR, each with a large number of inidividual checks - e.g. our smoke run has > 10 checks reported, our long run about 5. They are a mix of test results, warnings, and coverage, plus some custom checks. They are also duplicated _between_ parent jobs, for instance our build stage runs on both the long run and the deploy run.

Therefore, to help keep some kind of order to the large number of checks, we namespace them by the parent job, so we end up with:

```
Deploy  # github status, links to jenkins
Deploy / Build  # github check, links to github
Deploy / Deploy
Smoke
Smoke / Linting
Smoke / Unit Tests
Long
Long / Build
Long / Long Tests
```

etc.

The easiest way to achieve this, by far, is for every plugin that publishes a check to allow the user to provide the name of that check.

I'm also hoping that at some point the checks API will support _updating_ checks, so that we can have a pattern that looks like:

```
withCheck('My Check Name') { check ->
    sh 'test --output output.xml'
    junit testResults: 'output.xml', check: check
}
```

ie, a wrapper that creates a check in progress, catches and reports exceptions.

---

Having now seen the proposal - this would be totally unacceptable as a replacement. We sometimes invoke junit multiple times per stage, so would have to totally rearchitect our jenkinsfiles to meet this reduced flexibility.

Again, I think this would make a lot of sense as a default, but the ability to override must be kept.

mrginglymus:
[...]

PR #211, opened 2020-11-18
Default checks name to be ' / '-joined enclosing blocks names

mrginglymus:
This hopefully goes some way to address @jglick's concerns over the current behaviour (https://github.com/jenkinsci/junit-plugin/issues/210).

The default now (hopefully!) is that checks are published per junit invocation with a name derived from the enclosing blocks, unless otherwise overridden. This means that, assuming the idiom of one junit invocation per stage is followed, checks will not be overwritten losing data.

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -42,24 +42,26 @@ import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Publisher;
 import hudson.tasks.Recorder;
 import hudson.tasks.junit.TestResultAction.Data;
-import io.jenkins.plugins.junit.checks.JUnitChecksPublisher;
-import io.jenkins.plugins.junit.storage.FileJunitTestResultStorage;
-import io.jenkins.plugins.junit.storage.JunitTestResultStorage;
 import hudson.tasks.test.PipelineTestDetails;
 import hudson.util.DescribableList;
 import hudson.util.FormValidation;
+import io.jenkins.plugins.junit.checks.JUnitChecksPublisher;
+import io.jenkins.plugins.junit.storage.FileJunitTestResultStorage;
+import io.jenkins.plugins.junit.storage.JunitTestResultStorage;
+import jenkins.tasks.SimpleBuildStep;
+import org.apache.commons.collections.iterators.ReverseListIterator;
+import org.apache.commons.lang.StringUtils;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.types.FileSet;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 
+import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
-import javax.annotation.Nonnull;
-import jenkins.tasks.SimpleBuildStep;
-import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Generates HTML report from JUnit test result XML files.
@@ -282,7 +284,12 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
             }
 
             if (!task.isSkipPublishingChecks()) {
-                new JUnitChecksPublisher(build, task.getChecksName(), result, summary).publishChecks(listener);
+                // If we haven't been provided with a checks name, and we have pipeline test details, set the checks name
+                // to be a ' / '-joined string of the enclosing blocks names. JUnitChecksPublisher will handle defaults if
+                // checksName ends up being empty or null
+                String checksName = task.getChecksName() != null || pipelineTestDetails == null ? task.getChecksName() :
+                      StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
+                new JUnitChecksPublisher(build, checksName, result, summary).publishChecks(listener);
             }
 
             return summary;
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -28,6 +28,7 @@ import hudson.AbortException;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
@@ -285,10 +286,15 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
 
             if (!task.isSkipPublishingChecks()) {
                 // If we haven't been provided with a checks name, and we have pipeline test details, set the checks name
-                // to be a ' / '-joined string of the enclosing blocks names. JUnitChecksPublisher will handle defaults if
-                // checksName ends up being empty or null
-                String checksName = task.getChecksName() != null || pipelineTestDetails == null ? task.getChecksName() :
-                      StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
+                // to be a ' / '-joined string of the enclosing blocks names. If all of that ends up being empty or null,
+                // set a default of 'Test'
+                String checksName = task.getChecksName();
+                if (checksName == null && pipelineTestDetails != null) {
+                    checksName = StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
+                }
+                if (Util.fixEmpty(checksName) == null) {
+                    checksName = "Test";
+                }
                 new JUnitChecksPublisher(build, checksName, result, summary).publishChecks(listener);
             }
 
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -293,7 +293,7 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
                     checksName = StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
                 }
                 if (Util.fixEmpty(checksName) == null) {
-                    checksName = "Test";
+                    checksName = "Tests";
                 }
                 new JUnitChecksPublisher(build, checksName, result, summary).publishChecks(listener);
             }
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -287,7 +287,7 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
             if (!task.isSkipPublishingChecks()) {
                 // If we haven't been provided with a checks name, and we have pipeline test details, set the checks name
                 // to be a ' / '-joined string of the enclosing blocks names. If all of that ends up being empty or null,
-                // set a default of 'Test'
+                // set a default of 'Tests'
                 String checksName = task.getChecksName();
                 if (checksName == null && pipelineTestDetails != null) {
                     checksName = StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -61,6 +61,7 @@ import org.kohsuke.stapler.QueryParameter;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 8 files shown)


src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
```diff
@@ -1,4 +1,4 @@
 <div>
     If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
-    SCM hosting platforms. If not, a default of "Tests" will be used.
+    SCM hosting platforms. If not, a default comprising the enclosing blocks names will be used.
```
timja, line 3: how about? enclosing blocks is an implementation detail not meant for a non developers of this plugin
```suggestion
    SCM hosting platforms. If not, a default including the current stage / branch names will be used.
```

src/main/resources/hudson/tasks/junit/pipeline/JUnitResultsStep/help-checksName.html
```diff
@@ -1,4 +1,4 @@
 <div>
     If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
-    SCM hosting platforms. If not, a default of "Tests" will be used.
+    SCM hosting platforms. If not, a default comprising the enclosing blocks names will be used.
```
timja, line 3:
```suggestion
    SCM hosting platforms. If not, a default including the current stage / branch names will be used.
```

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -282,7 +284,12 @@ public static TestResultSummary parseAndSummarize(@Nonnull JUnitTask task, Pipel
             }
 
             if (!task.isSkipPublishingChecks()) {
-                new JUnitChecksPublisher(build, task.getChecksName(), result, summary).publishChecks(listener);
+                // If we haven't been provided with a checks name, and we have pipeline test details, set the checks name
+                // to be a ' / '-joined string of the enclosing blocks names. JUnitChecksPublisher will handle defaults if
+                // checksName ends up being empty or null
+                String checksName = task.getChecksName() != null || pipelineTestDetails == null ? task.getChecksName() :
```
timja, line 290: example PR where this has been tested?

mrginglymus: I'm going to spend some quality time with mockito to see if I can get the checks tests testing more of the implementation of `parseAndSummarize`.

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -2,29 +2,40 @@
...
+import org.powermock.modules.junit4.PowerMockRunner;
 
 import static java.util.Objects.requireNonNull;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertNotNull;
+import static org.mockito.ArgumentMatchers.*;
+import static org.powermock.api.mockito.PowerMockito.*;
 
+@RunWith(PowerMockRunner.class)
```
jglick, line 25: Beware that every time I have come across Mockito or PowerMock tests in Jenkins core or plugins, it has ended in disaster. They tend to require way more maintenance than `JenkinsRule` tests (with `@TestExtension` mocks as needed), and fail to detect real regressions.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -282,7 +284,12 @@ public static TestResultSummary parseAndSummarize(@Nonnull JUnitTask task, Pipel
...
+                      StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
```
jglick, line 291: Are the spaces around the slash conventional?

jglick, line 291: What if `enclosingBlockNames` is empty? (Possible, in fact common, in Scripted syntax.)

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
mrginglymus: It has been fun getting them to work (mostly because my work machine decided to OOM every other run...) but I couldn't see another obvious point at which to intercept this _without_ some kind of mocks. I'm certainly not tied to the current implementation, but we have improved coverage this way at least.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
mrginglymus: I took my inspiration from https://github.com/jenkinsci/junit-plugin/blob/a39b176d80f8e984b9c8ec7ac22c4b2ac1d41b93/src/main/java/hudson/tasks/junit/CaseResult.java#L319

mrginglymus: That should result in an empty string being passed in, which will be fixed in the `JUnitChecksPublisher` ctor. Perhaps it would make more sense to stick the default here instead. Either way, this scenario should now be tested.

jglick: Right…this is a display name for rendering in an HTML GUI, though, whereas a check name is in a somewhat different category because it is used as a key in the REST API, requested in branch protection settings, etc.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-checksName.html
```diff
@@ -1,4 +1,4 @@
 <div>
     If provided, and publishing checks enabled, the plugin will use this name when publishing results to corresponding
-    SCM hosting platforms. If not, a default of "Tests" will be used.
+    SCM hosting platforms. If not, a default including the current stage / branch names will be used.
```
jglick, line 3: This build step would not normally be used in Pipeline jobs. (It is still _possible_, via some rather gritty Groovy syntax, because @abayer left this as a `SimpleBuildStep` after introducing `JUnitResultsStep`, but very unlikely.) So this help would be displayed to users of traditional job types, for which “stage” and “branch” are meaningless.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
mrginglymus: I've been using slashes in my own custom checks without issue; nothing seems to mind.

>  used as a key in the REST API

As in in Githubs REST API?

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
jglick: > obvious point at which to intercept this

Not sure I am following. https://github.com/jenkinsci/junit-plugin/blob/91f112be2b2b7f9bc056eb7ee0bb4f6ab7d645c2/src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java#L32-L40 and the like already run actual builds with stages and actual test results, and https://github.com/jenkinsci/junit-plugin/blob/91f112be2b2b7f9bc056eb7ee0bb4f6ab7d645c2/src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java#L51 already asserts the check name.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
jglick: > I've been using slashes in my own custom checks without issue

Yes of course, I am just asking if there is a convention for using strings that look more like display names (`Tests / Branch / Stage`) or more like file paths or URIs (`tests/branch/stage`).

> As in in Githubs REST API?

Yes.

mrginglymus: > or more like file paths or URIs (tests/branch/stage).

I'm not sure I see where.

> file paths or URIs (tests/branch/stage).

ahh, to look more like this?

image

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
mrginglymus: That only tests the constructor of `JUnitChecksPublisher`; without this change none of the code in `JUnitResultArchiver#parseAndSummarize` is tested, which is where we have access to the task config and pipline test details. See for example 
https://github.com/jenkinsci/junit-plugin/blob/91f112be2b2b7f9bc056eb7ee0bb4f6ab7d645c2/src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java#L138
Where we can't (currently) test that the custom check name is correctly extracted. 

This new way of testing would also allow us to write tests for the bug in https://github.com/jenkinsci/junit-plugin/pull/199, which again was code in `parseAndSummarize` that did not have coverage.

timja: you could extract that code out to somewhere easier to test, either in the checks publisher or somewhere else?

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
timja: I prefer display name. something like `Tests / Unit`

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
mrginglymus: I'll have a play :)

mrginglymus: Ahh, thanks for the hint on `@TestExntesion`, that's much neater.

Apologies if my implementation here isn't terribly idiomatic; if it's not already obvious my Java is a bit rusty and I'm not really up to date with the best ways of doing things.

jglick: I am afraid Jenkins programming has little to do with general Java experience; it is very idiosyncratic due to the old bespoke frameworks.

jglick (commented):
Looking good!

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -1,39 +1,72 @@
...
+        @Override
+        protected Optional<ChecksPublisher> createPublisher(Job<?, ?> job, TaskListener listener) {
+            return Optional.of(publisher);
+        }
+    }
+
+    @After
+    public void clearDetails() {
+        InterceptingChecksPublisherFactory.publisher.details.clear();
+    }
```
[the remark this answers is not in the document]
mrginglymus: Ahh, thanks! I figured there must be something like that.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -282,7 +285,17 @@ public static TestResultSummary parseAndSummarize(@Nonnull JUnitTask task, Pipel
             }
 
             if (!task.isSkipPublishingChecks()) {
-                new JUnitChecksPublisher(build, task.getChecksName(), result, summary).publishChecks(listener);
+                // If we haven't been provided with a checks name, and we have pipeline test details, set the checks name
+                // to be a ' / '-joined string of the enclosing blocks names. If all of that ends up being empty or null,
+                // set a default of 'Tests'
+                String checksName = task.getChecksName();
```
timja, line 291: shouldn't we still include tests in the check ?

image

mrginglymus: Hmm, yes. For context I'm working mostly on python/js environments, where our stages are dedicated to test generation...

```groovy
stage('JS Tests') {
    sh 'yarn test'
    junit 'js_tests.xml'
}

stage('Python Tests') {
    sh 'pytest tests/'
    junit 'python_tests.xml'
}
```

...as opposed to maven where tests are a bit more of a 'side effect' of the build

```groovy
stage('Build') {
    sh 'mvn install'
    junit 'results.xml'
    archiveArticfcats(...)
    recordIssues(...)
    // etc...
}
```

So we'd end up with `JS Tests / Tests` etc. OTOH, I'm already setting a custom name (and will be using `withChecks` when it's ready) so I think meeting the 'mvn' use case possibly makes more sense?

timja: Yeah I think so

timja: I made the change locally,

```
+
+    private static final String DEFAULT_CHECKS_NAME = "Tests";

     /**
      * {@link FileSet} "includes" string, like "foo/bar/*.xml"
@@ -290,10 +294,12 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
                 // set a default of 'Tests'
                 String checksName = task.getChecksName();
                 if (checksName == null && pipelineTestDetails != null) {
-                    checksName = StringUtils.join(new ReverseListIterator(pipelineTestDetails.getEnclosingBlockNames()), " / ");
+                    List<String> checksComponents = new ArrayList<>(pipelineTestDetails.getEnclosingBlockNames());
+                    checksComponents.add(DEFAULT_CHECKS_NAME);
+                    checksName = StringUtils.join(new ReverseListIterator(checksComponents), " / ");
                 }
                 if (Util.fixEmpty(checksName) == null) {
-                    checksName = "Tests";
+                    checksName = DEFAULT_CHECKS_NAME;
                 }
```

Requires tests changes but that gives what I expected

mrginglymus: OK, will get that done

mrginglymus: Ugh, sorry, you meant `Tests` at the beginning...

timja:
@mrginglymus 

🎉 https://github.com/jenkinsci/bom/pull/374

image

timja:
https://github.com/jenkinsci/junit-plugin/issues/221 May be related, I didn’t do any complex testing in my setup

PR #220, opened 2020-12-03
Fix NPE error

nirkoren:
Hi guys, I think it might improve the code.. Don't you? 👍


timja (changes requested):
spotbugs says no, this is redundant

Issue #222, opened 2020-12-04
MissingFieldException 'checksName' for Freestyle project


zencode1:
After upgrading JUnit plugin to 1.44 Jenkins warns of unreadable data for freestyle projects that use publish JUnit results.

```
Project type: hudson.model.FreeStyleBuild
Error: MissingFieldException: No field 'checksName' found in class 'hudson.tasks.junit.TestResultAction
```

Since I don't use any custom names I chose to discard the old data.  The same day, I upgraded JUnit to 1.45.  Unrelated but I mention this just in case, I was finally able to upgrade Jenkins past 2.263 (couldn't due to the div vs tables change).

This morning, the unreadable data messages are shown again.  This time, shown using Jenkins 2.269 formatting.

```
Project type:
Error: MissingFieldException: Field not found in class. ---- Debugging information ---- message : Field not found in class. field : hudson.tasks.junit.TestResultAction.checksName
```

I clicked "Discard Unreadable Data".  Is this going to continue to happen every day?

timja:
If I knew how to reproduce it I could fix it, but I just can't seem to reproduce this on demand.

(We have this happening in our production instance too)

I can't see why `checksName` is expected to be on that class =/
I'm misunderstanding some Jenkins-ism I think. but can't see what

> I clicked "Discard Unreadable Data". Is this going to continue to happen every day?

for now yes probably, shouldn't cause an actual issue though

timja:
Looks to be caused by https://github.com/jenkinsci/junit-plugin/pull/199/files#diff-4fa8f2162a640cc3f024dadbffc9b442ab73e2afd30e5030ec0088ac300371e5L78

This should go away if you reconfigure your job

PR #231, opened 2020-12-21
Support withChecks context

mrginglymus:


src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
```diff
@@ -48,7 +52,14 @@ protected TestResultSummary run() throws Exception {
         pipelineTestDetails.setNodeId(nodeId);
         pipelineTestDetails.setEnclosingBlocks(getEnclosingBlockIds(enclosingBlocks));
         pipelineTestDetails.setEnclosingBlockNames(getEnclosingBlockNames(enclosingBlocks));
+
         try {
+            // If we are within a withChecks context, and have not provided a name override in the step, apply the withChecks name
+            if (Util.fixEmpty(step.getChecksName()) == null) {
+                Optional.ofNullable(getContext().get(ChecksInfo.class))
+                        .map(ChecksInfo::getName)
+                        .ifPresent(step::setChecksName);
```
timja, line 61: I was wondering if we should be deprecating the explicit setting of the checks name?

going forward I wouldn't expect we would need to add these overrides to steps as we have `withChecks` what do you think?

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -232,4 +238,71 @@ public void extractChecksDetailsEmptySuite() throws Exception {
...
     }
+
+    @Test
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
+    public void withChecksContext() throws Exception {
+        WorkflowJob j = rule.jenkins.createProject(WorkflowJob.class, "singleStep");
+        j.setDefinition(new CpsFlowDefinition("stage('first') {\n" +
+                "  node {\n" +
+                "    withChecks('With Checks') {\n"+
+                "      def results = junit(testResults: '*.xml')\n" +
```
timja, line 249:
```suggestion
                "      def results = junit '*.xml'\n" +
```

timja (approved):

timja:
I assume draft is wait on the checks api release?

timja (commented):
Can you update the README to please? At the bottom

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
mrginglymus: It makes a nice shorthand - for instance

```groovy
withChecks('Maven build') {
    sh 'maven clean install'
    publishChecks()
}
junit testResults: '*.xml', checksName: 'Custom tests name'
```

rather than

```groovy
withChecks('Maven build') {
    sh 'maven clean install'
    publishChecks()
}
withChecks('Custom tests name') {
    junit testResults: '*.xml'
}
```

but to be honest, that second one doesn't look to bad, and I'm struggling to find a use case where you'd actually want to override a `withChecks` setting. I certainly can't think of one within the suite of Jenkinsfiles I manage.

timja: I guess it's needed for declarative pipeline? the override that is, until we ship a compatible version for it.

unless withChecks already works for that?

mrginglymus: Apparently it Just Works (tm)  - https://github.com/jenkinsci/junit-plugin/pull/231/commits/c2e320810de425f726082a7494a2e0225845505c

timja: hmm ok, I guess we leave it in for now although it does cause this somehow:
https://github.com/jenkinsci/junit-plugin/issues/222

Which I haven't managed to figure it out

README.md
```diff
@@ -53,6 +53,14 @@ In order to disable the checks feature, set the property `skipPublishingChecks`
 junit skipPublishingChecks: true, testResults: 'test-results.xml'
 ```
 
+The plugin will use a checks name provided by a `withChecks` context.
```
timja, line 56:
```suggestion
The plugin will default to using the stage name or branch of a parallel step for the checks name, if those aren't present 'Tests' will be used, the name can be overridden by a `withChecks` step.
```

README.md
```diff
@@ -53,6 +53,14 @@ In order to disable the checks feature, set the property `skipPublishingChecks`
...
+
+```groovy
+withChecks('Integration Tests') {
+  junit 'test-results.xml'
+}
+```
+
 In order to provide a custom check name, set the property `checksName`:
```
timja, line 64: I think we can make this undocumented now and recommend `withChecks`?

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
mrginglymus: Was that caused by https://github.com/jenkinsci/junit-plugin/pull/199/files#diff-4fa8f2162a640cc3f024dadbffc9b442ab73e2afd30e5030ec0088ac300371e5L78 ?

timja:
enforcer issues since adding the declarative pipeline test, probably just exclude the ancient version of joda-time

src/main/java/hudson/tasks/junit/pipeline/JUnitResultsStepExecution.java
timja: ah yeah probably, I guess it'll go away as jobs are re-configured? unless we want to add the field back, bit annoying though

timja (approved):

PR #236, opened 2021-01-10
Use truncated strings from checks api

mrginglymus:
Consumer implementation of https://github.com/jenkinsci/checks-api-plugin/pull/68

Is there a way we can set a constraint on the version of `github-checks-plugin` without requiring it to be installed, such that we don't have to go through a gradual deprecation process of the old logic?


timja (approved):
looks sane, I wrote a test for this initially but I couldn't manage to make windows CI pass and I think I ended up deleting the test

timja (approved):

timja:
> Is there a way we can set a constraint on the version of github-checks-plugin without requiring it to be installed, such that we don't have to go through a gradual deprecation process of the old logic?

Short of putting it in the release notes, no

timja (approved):

PR #272, opened 2021-05-21
Upgrade trends and detail views to Bootstrap 5

uhafner:
Bootstrap 5 has been released and requires some changes in markup and JS. Additionally, (due to the migration of ECharts to Bootstrap 5) trend charts are now configurable using a dialog.

Upstream PR: jenkinsci/echarts-api-plugin#124

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>echarts-api</artifactId>
-            <version>5.1.0-3-rc454.dbbdd5633ce2</version>
+            <version>5.1.0-3-rc464.d7d9d5650d0a</version>
         </dependency>
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>bootstrap5-api</artifactId>
-            <version>5.0.1-1</version>
+            <version>5.0.1-2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>echarts-api</artifactId>
-            <version>5.1.0-3-rc443.de651a379c4a</version>
+            <version>5.1.0-3-rc454.dbbdd5633ce2</version>
         </dependency>
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,12 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-project</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <artifactId>junit</artifactId>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -201,7 +207,7 @@
             <dependency>
                 <groupId>io.jenkins.plugins</groupId>
                 <artifactId>font-awesome-api</artifactId>
-                <version>5.15.3-2</version>
+                <version>5.15.3-3</version>
             </dependency>
             <dependency>
                 <groupId>io.jenkins.plugins</groupId>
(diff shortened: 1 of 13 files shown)


src/main/java/hudson/tasks/junit/History.java
```diff
@@ -66,46 +69,60 @@ public boolean historyAvailable() {
         return testObject.getRun().getParent().getBuilds().size() > 1;
     }
 
-    @JavaScriptMethod
+    @Deprecated
```
timja, line 72: just remove it, it's a web method only

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -66,46 +69,60 @@ public boolean historyAvailable() {
...
                 return new TestResultTrendChart().create(pluggableStorage.getTrendTestResultSummary());
             }
         }
 
-        return new TestResultTrendChart().createFromTestObject(createBuildHistory(testObject), new ChartModelConfiguration());
+        return new TestResultTrendChart().createFromTestObject(createBuildHistory(testObject), chartModelConfiguration);
     }
 
-    @JavaScriptMethod
+    @Deprecated
```
timja, line 96: remove also

uhafner:
Seems that HTMLUnit does not understand modern JS :-( Browser and IntelliJ do not complain. This is one of the reasons why I removed HTMLUnit tests from my plugins. They break very often due to incompatibilities (and are soooo slow).

```
EChartsJenkinsApi.prototype.readConfiguration = function (id) {
    const specific = echartsJenkinsApi.readFromLocalStorage(id);
    const common = echartsJenkinsApi.readFromLocalStorage(trendDefaultStorageId);

    return {
        ...specific,
        ...common
    };
}
```

timja:
Everything in this plugin is inherited, probably from when it was split from core. 

I’ll try take a peek at this soon and see if I can find a way forward

jtnord (commented):
I thought there was a trend toward getting rid of shared js libs not adding more  (using webpack and creating isolated bundles for each plugin).

If this is CSS are we going to have collisions where BS 4 and BS 5 are in the same page
@EstherAF

timja:
I think only the history page uses bootstrap and that will be explicitly loading bootstrap 4

timja:
Now failing for https://github.com/HtmlUnit/htmlunit/issues/232

uhafner:
I have a testing course in the winter semester again. If it is helpful we could include the JUnit plugin as project and restructure it so that the ATH tests will run inside the plugin build as well. There is already a test at https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/JUnitPluginTest.java that just needs to be moved (and maybe extended). 

As already mentioned above, I also think using HTMLUnit In general is not very helpful as it is just a (slow and buggy) simulation of a real browser. It works quite well with the static HTML pages of the old style Jenkins UI but not with modern pages that use a lot of other JS libraries. I think it make more sense if we would use unit and integration tests only to verify the Java models and not the browser rendering. Browser render should only be tested (if at all) using ATH.

timja:
> I have a testing course in the winter semester again. If it is helpful we could include the JUnit plugin as project and restructure it so that the ATH tests will run inside the plugin build as well. There is already a test at [https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/JUnitPluginTest.java](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/JUnitPluginTest.java?rgh-link-date=2021-08-09T07%3A13%3A54Z) that just needs to be moved (and maybe extended).

If you could do that it would be great!

Absolutely onboard with the conversion of these to ATH tests.

(In the meantime if rhino adds support for what we need we'll be able to ship this)

pom.xml
```diff
@@ -48,16 +48,7 @@
...
-        <dependency>
-            <groupId>io.jenkins.plugins</groupId>
-            <artifactId>checks-api</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.jenkins.plugins</groupId>
-            <artifactId>checks-api</artifactId>
-            <scope>test</scope>
-            <classifier>tests</classifier>
+            <version>5.1.2-6-rc504.fd62772631c3</version>
```
timja, line 51:
```suggestion
            <version>5.1.2-6</version>
```

timja (approved):
good enough with the limitations we have, I'm not sure if the tests are actually testing much now given the endpoints aren't being called with JS disabled but let's get this in so we can move forward in bom

timja:
I've pushed origin branches:

https://github.com/jenkinsci/junit-plugin/tree/1.53.x
https://github.com/jenkinsci/junit-plugin/tree/1.54.x

I spent sometime trying to backport it but found it difficult, lots of pom issues.

Feel free to submit PRs to those branches, otherwise probably won't do? I believe you have found a workaround in bom

basil:
I somehow managed to find a way to work around this for now, but …

> otherwise probably won't do

… in general, if you aren't willing to do backports, then I suggest not increasing the baseline so aggressively.

PR #286, opened 2021-08-08
Add color configuration of trend chart (show passed tests in green or blue )

uhafner:
The new configuration option allows choosing green or blue as color for passed tests:

Bildschirmfoto 2021-08-08 um 23 42 33

Additionally, the stacked chart series have been reordered so that the passed tests come first.
Then the automatic axis scaling of ECharts works as expected and shows only the important chart part. In the screenshot below the x-Axis now starts at 500.

Here is a screenshot of the new layout:
Bildschirmfoto 2021-08-08 um 23 42 17

This PR is an add-on for #272. I will rebase and cleanup this PR after #272 has been merged.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,17 @@
+# Note: additional setup is required, see https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc
+
+name: Release Drafter
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
+      # Drafts your next Release notes as Pull Requests are merged into "master"
+      - uses: release-drafter/release-drafter@v5
+        env:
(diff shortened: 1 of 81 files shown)


timja:
Nice we could probably change the default colour to green?

uhafner:
> Nice we could probably change the default colour to green?

Yes, absolutely. This makes more sense for JUnit results (and the build success status is green anyway).

timja:
Some conflicts from the history bug fixes btw.

uhafner:
Yes, I also need to clean up the patch, I forgot to disable IntelliJ's new on save actions. I will do this after the #272 has been merged.

timja:
> Yes, I also need to clean up the patch, I forgot to disable IntelliJ's new on save actions. I will do this after the #272 has been merged.

sure, unlikely code will change much in this area now so any patch should be stable

timja (approved):
What do think about swapping the default?

uhafner:
I have no strong opinion. Swapping the default would not change the existing visualization, so it might be a better start.

timja:
Ah actually this is the right way round, I just tested it out. Looks good

PR #342, opened 2022-01-20
Feature/UI tests

NikolasFunction:
Added ui tests for the junit plugin.
https://issues.jenkins.io/browse/JENKINS-67275


NikolasFunction:
@jglick @kohsuke @timja 

Could one of you please add @uhafner as a PR reviewer?
It seems he doesn't have rights for this action. Thanks!

jtnord:
What is the reasoning to do this here?
UI testing will be brittle with changes in Jenkins core so the BOM will likely be failing regularly for weeklies, and or versions held back from the BOM for some lines.

jglick:
Note that this will complicate infrastructure somewhat. https://github.com/jenkinsci/plugin-compat-tester/pull/249

uhafner:
> What is the reasoning to do this here?

I think UI tests belong to a plugin like unit and integration tests (the idea is from @olivergondza, see https://github.com/jenkinsci/acceptance-test-harness/pull/556 for details). After contributing to ATH for several years now I made the observation that nobody is actually looking at UI tests that are located somewhere else. So it makes sense to have them in the plugin. Then we see if a new feature breaks an existing feature (on the UI side) or if an upgrade to a more recent core version breaks the UI.

Additionally, JUnit plugin uses integration tests with HTMLUnit which does not support modern JS frameworks (see #272, #286) . So it makes sense to remove those tests in favor of real UI tests. 

> UI testing will be brittle with changes in Jenkins core so the Jenkins tools BOM will likely be failing regularly for weeklies, and or versions held back from the BOM for some lines.

Why should that be a problem? If the tests will break then something in the UI is wrong. That means we should fix it.

uhafner:
> Note that this will complicate infrastructure somewhat. [jenkinsci/plugin-compat-tester#249](https://github.com/jenkinsci/plugin-compat-tester/pull/249)

I haven't looked in detail at those changes but I would assume that we would have two alternatives: a single module plugin and a multi-module plugin. So the `ifWarnings` blocks should be actually `ifMultiModule` blocks. 

Would it be possible to just use the plugin module in those tests and ignore the other ones?

timja:
As Ulli said the critical reason for this is modern JS frameworks don't work with HTMLUnit 😢

uhafner (changes requested):
See comments in https://github.com/jenkinsci/junit-plugin/pull/355#pullrequestreview-888093560. 

GitHub is unable to render this PR in my browser.

Main review comments
- Use the correct AssertJ assertions
- Don't duplicate login in page objects (2 constructors)

It would be also helpful if you can include one additional `SmokeTest` that exercises the most important UI elements in one row so that this test can be used during PR builds. (The other UI tests can be then used only as additional tests).

jtnord (commented):
I forgot to finish my review :(

uhafner:
@NikolasFunction @Tjuri Are you still planning to update this PR? The longer you are waiting the more merge conflicts will occur...

timja:
Principle is good but this has been abandoned

Issue #344, opened 2022-01-30
Test view table has broken dropdowns


josephjthomas:
### Jenkins and plugins versions report

Jenkins 2.332, JUnit plugin 1.53

### What Operating System are you using (both controller, and any agents involved in the problem)?

OS: Linux - 3.10.0-1160.45.1.el7.x86_64

### Reproduction steps

1. Install the JUnit plugin
2. Archive tests for a project that has JUnit tests
3. Head to the `/testReport/` tab of a build

### Expected Results

I am not sure which kind of information used to be displayed here, maybe it would make sense to remove the drop down button overall to avoid confusion.

### Actual Results

All tests have a non working dropdown. The same applies to the `/testReport/history/` tab.
![](https://i.imgur.com/ywhwQyc.png)

timja:
PR that fixes it is up at https://github.com/jenkinsci/junit-plugin/pull/345, I've asked a couple of people if they know if it's supposed to be used for something

PR #345, opened 2022-01-30
Remove arrow for dropdown that does nothing

timja:
Fixes https://github.com/jenkinsci/junit-plugin/issues/344

### DIFF
--- a/src/main/resources/hudson/tasks/junit/CaseResult/list.jelly
+++ b/src/main/resources/hudson/tasks/junit/CaseResult/list.jelly
@@ -41,7 +41,7 @@ THE SOFTWARE.
         <j:if test="${test != null}">
           <tr>
             <td class="pane">
-              <a href="${app.rootUrl}${b.url}" class="model-link inside">${b.fullDisplayName}</a>
+              <a href="${app.rootUrl}${b.url}">${b.fullDisplayName}</a>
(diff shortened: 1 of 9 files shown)


timja:
@Vlatombe or @jglick any chance you have an idea if this is actually supposed to be used for something? (see https://github.com/jenkinsci/junit-plugin/issues/344)

jglick:
`class="model-link inside"` is supposed to be used for `ModelObject`s which have a context menu. At least `Item`s, `Run`s, etc. have this. In this case it seems `https://jenkins/job/xxx/lastSuccessfulBuild/testReport/some.pkg/contextMenu` is a 404 and ditto for other test objects. No idea if these _used_ to have a context menu prior to some refactoring or what. I do not see anything interesting in the sidebar (just the same links as on the `Run`) so perhaps not. If you were curious you could dig through Git history to find out why this CSS class was initially added.

timja:
https://github.com/jenkinsci/junit-plugin/commit/4a3d5130394b1c549dc25e87752546468ed74bbb

timja:
Actually looks like a mistake in https://github.com/jenkinsci/junit-plugin/commit/c21c787c32d858767c536bae14872d790bc43d9f

no reason given why they were converted to model-link

jglick:
Yup, just got there myself. :shrug: go ahead as far as I am concerned.

PR #347, opened 2022-02-01
Utilize new table style

NotMyFault:
The change proposed adapts the new table style provided by core.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <revision>1.54</revision>
         <changelist>-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-        <jenkins.version>2.303.3</jenkins.version>
+        <jenkins.version>2.321</jenkins.version>
         <java.level>8</java.level>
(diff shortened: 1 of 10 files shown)


timja (commented):
Doesn’t this need the minimum version of core increased too?

NotMyFault:
> Doesn’t this need the minimum version of core increased too?

You are right. I, for some reason, was under the impression this change used to be out for much longer. Probably better to keep this on hold then, the next LTS candidate would be the first one to include it anyway.

timja:
There’s no need to hold it back, just update the minimum version to the one it was released in

NotMyFault:
> There’s no need to hold it back, just update the minimum version to the one it was released in

That works too, I wasn't too sure whether that was desired or not.

timja (approved):

uhafner:
In Jenkins 2.345 the tables have now a rectangular design, is this intended?

Screenshot:

Bildschirmfoto 2022-05-02 um 18 56 51

timja:
I don't think so

NotMyFault:
That indeed is the wrong layout for this kind of table 👀

Issue #201, opened 2020-10-29
Allow old test reports


ZakTaccardi:
I'm seeing an incorrect error:

```
Recording test results
Test reports were found but none of them are new. Did leafNodes run? 

For example, /home/tools/jenkins/workspace/../mySdk/build/test-results/testDebugUnitTest/TEST-com.example.ExampleTest.xml is 16 hr old
```

This is correct, because I am using Gradle and it has identified that it did not need to execute the tests because it was previously run and no inputs have changed from a previous build

Some of my code:

```groovy
@Field String jUnitReports = [
        '**/build/test-reports/*.xml',
        '**/build/test-results/**/*.xml',
        '**/build/outputs/androidTest-results/**/*.xml',
        '**/spoon-output/**/junit-reports/*.xml',
        '**/junit-reports/*.xml',
        '**/build/outputs/androidTest-results/**/*.xml',
]
        .join(",")
```

```groovy
post {
  always { 
    junit testResults: jUnitReports, allowEmptyResults: true
}
```

Can a flag be added to allow old test reports and skip this check? Maybe `allowOldTestReports: true`?

Note: currently using JUnit Plugin version 1.28

Relates to:
- https://issues.jenkins-ci.org/browse/JENKINS-47315
- https://issues.jenkins-ci.org/browse/JENKINS-6268
- https://github.com/jenkinsci/junit-plugin/pull/108

timja:
There's a PR for it here: https://github.com/jenkinsci/junit-plugin/pull/108

mpress:
> There's a PR for it here: #108

A PR that is over two years old :(

pappnase99:
I'm seeing the same error, although my test results are only "11 sec old". Test is flaky, sometimes successful, but mostly failing. No idea what to do, the related PR is closed.

liyaka:
we are in 2022 and still the same problem - is there anything that can be done??

timja:
submit a PR for it?

timja:
Also having someone tell me:

> so do it yourself

Is a great way to have this drop to the bottom of my priorities =/

Yohandah:
I think rejecting 2 PR fixing this issue and the fact it's been 4 years: it's already at the bottom of ur list ! It doesn't encourage ppl to contribute either. 
No worries, there's workarounds

jtnord:
NB: could be reverted as this changes the behaviour of the plugin from reporting tests that ran to any test there is a report for.
[...]

PR #348, opened 2022-02-02
Don't error on old test reports

timja:
This is an outdated feature that doesn't reflect modern build tools that won't rebuild if all the inputs are the same (and they aren't forced to rebuild)

A multitude of hacks for gradle that are no longer needed after this:
https://stackoverflow.com/questions/13879667/how-to-fix-test-reports-were-found-but-none-of-them-are-new-did-tests-run-in

Seems like there was no test coverage for this and I don't think it makes sense to add any for a non feature

Fixes https://github.com/jenkinsci/junit-plugin/issues/201

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResult.java
+++ b/src/main/java/hudson/tasks/junit/TestResult.java
@@ -216,32 +216,11 @@ public final class TestResult extends MetaTabulatedResult {
      */
     public void parse(long buildTime, File baseDir, PipelineTestDetails pipelineTestDetails, String[] reportFiles) throws IOException {
 
-        boolean parsed=false;
 
         for (String value : reportFiles) {
             File reportFile = new File(baseDir, value);
             // only count files that were actually updated during this build
-            if (buildTime-3000/*error margin*/ <= reportFile.lastModified()) {
-                parsePossiblyEmpty(reportFile, pipelineTestDetails);
-                parsed = true;
-            }
-        }
-
-        if(!parsed) {
-            long localTime = System.currentTimeMillis();
-            if(localTime < buildTime-1000) /*margin*/
-                // build time is in the the future. clock on this agent must be running behind
-                throw new AbortException(
-                    "Clock on this agent is out of sync with the controller, and therefore \n" +
-                    "I can't figure out what test results are new and what are old.\n" +
-                    "Please keep the agent clock in sync with the controller.");
-
-            File f = new File(baseDir,reportFiles[0]);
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
@timja I believe this change broke https://github.com/jenkinsci/maven-plugin/blob/0f600837dbea3d8a86984f34f9cc7eb2e24950da/src/test/java/hudson/maven/reporters/SurefireArchiverUnitTest.java#L175 which was created due to https://issues.jenkins.io/browse/JENKINS-31524 which I believe is still a valid concern.

My knowledge on the internals here is zero so I believe you are the best one to evaluate if JENKINS-31524 is still a concern or is no more relevant

timja:
Hi 

I've had a bit of a look and can't say I really understand the maven plugin change that was done.
I would've thought the change here would only help in the reported bug case but not the easiest to understand.

The safest way would be for the user to use separate directories anyway.

raul-arabaolaza:
Thanks @timja I am on the same situation, will dig deeper.

jtnord:
But if nothing is rebuilt, and hence the tests are not run, (and the file is outdated) why would we want to report the tests as run / passing in the UI when they are not run?

Because the tests are not run how do you know they pass.  For example a test that fails if the date > 19-03-2022  would pass today.
If gradle or whatever never re team that test as the production and test code never changes we would report it tomorrow as passing. (And it would not have passed if ran).
This leads to the issue that people would see a failing test in thier PR, (when compared to the last build on the mainline branch) which with good reason they could assume was due to thier PR.

Iirc (unable to check now) but this issue reported was always fixable if the user wanted it by setting a system property?

The issue we also have now IIUC is that we will also report tests that no longer exist.
E.g do not clean target/ before running tests and remove a Test file?

jtnord:
> This fundamentally did not work with gradle's test results previously without work arounds.

does it work now with `maven` (`mvn test`) for example then `rm -fr src/test/java/myTest`   (the surefire result for test `MyTest` will now remain and be constantly reported until you wipe out the workspace).

> This change means that the plugin reports test results that are there. If you don't want them to be reported then you need to not have them there.

And that is the issue - the plugin should report tests that *ran*

jglick:
This appears to leave behind an unused parameter `buildTime` as well as an obsolete comment https://github.com/jenkinsci/junit-plugin/blob/86b06fd8de7396664cd59cd7400b33414181537a/src/main/java/hudson/tasks/junit/TestResult.java#L142-L143 IIUC?

PR #350, opened 2022-02-10
Add UI tests

uhafner:
Temporary followup for #342 to see if all tests pass in CI with latest release of junit-plugin.


timja:
I touched it, maybe it'll work

timja:
I'm happy to give you write access for playing with this?

uhafner:
I don't want to use additional resources on Jenkins.ci :-)

uhafner:
But maybe it makes sense to have permissions then I can push in a branch here so that the actions will run. The setup is similar to the warnings plugin.

timja:
Gave you write access

PR #375, opened 2022-04-16
pin jackson2-api temporarily

jetersen:
bump jackson2-api to avoid cyclic dependency


jetersen:
We need to pin this to unblock https://github.com/jenkinsci/bom/pull/1011#issuecomment-1100675121

jetersen:
@timja do you mind reviewing / releasing this? 😄

PR #377, opened 2022-04-21
Update plugin POM and BOM

basil:


basil:
The changes to `TestResultStorageJunitTest` are because the newer version of Surefire has buffer overflow issues when tests write too much to standard output. This PR works around the problem by using a logger instead.

@timja A release of this would facilitate Java 17 PCT testing.

timja (approved):

basil:
Thank you!

Issue #368, opened 2022-03-22
plugin reports all existing tests, not tests that ran


jtnord:
### Jenkins and plugins versions report

the plugin now reports all tests, and not just tests that run in the build.

https://github.com/jenkinsci/junit-plugin/pull/348#issuecomment-1072274140

### What Operating System are you using (both controller, and any agents involved in the problem)?

all

### Reproduction steps

Build a project
wait an hour
delete all the tests
build the project again (in the same agent with the same workspace)

### Expected Results

no test results are reported as no tests ran

### Actual Results

all the tests that that previously ran reported.

lppedd:
Hi @jtnord! Any workaround for us fellas that are still on older releases of Jenkins/plugin?

PR #384, opened 2022-05-10
[WIP] restore not parsing of old test result files but as a configurable option

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>


src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -934,4 +968,14 @@ public CaseResult getCase(String suiteName, String transformedFullDisplayName) {
                 .findFirst()
                 .orElse(null);
     }
+
+    private transient PrintStream logger;
+
```
olamy, line 973: I definitely do not consider this as a really elegant way of logging :) but not sure what are the other options to get some debug/output in the console when running tests.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -56,4 +56,7 @@ THE SOFTWARE.
     <f:entry title="${%Skip marking build as unstable on test failure}" field="skipMarkingBuildUnstable">
             <f:checkbox default="false" title="${%If checked, the test failures will still be reported but won't mark the build as unstable}"/>
     </f:entry>
+    <f:entry title="${%Include report files older than build start}" field="parseOldReports">
+            <f:checkbox default="false" title="${%If checked, the test report files older than build start will be included in the parsing}"/>
```
timja, line 60: this implies that it is defaulting to the poor behaviour of before?

I expected the other way around, see https://github.com/jenkinsci/junit-plugin/pull/367#issuecomment-1122369873

olamy: I guess this behaviour just makes everybody happy.
Those who are using static nodes will have same behaviour as before and those who are using temporary node will not have problem neither.

timja: I haven't read the complete changeset but I guess what you're implying (not described in the PR):

There's no longer any error thrown that was thrown in the previous previous behaviour but it defaults to not showing 'cached' or 'left over' test results.

cc @jglick

olamy: > I haven't read the complete changeset but I guess what you're implying (not described in the PR):
> 
> There's no longer any error thrown that was thrown in the previous previous behaviour but it defaults to not showing 'cached' or 'left over' test results.
> 

any unit test for that?

timja: No as it was the absence of a feature it didn’t make sense to test that to me.

the previous behaviour wasn’t even tested in the plugin

olamy: well frankly if a majority think removing a long existing behaviour and/or change it as non default makes sense.
why not... I do not have strong argument. I just want to give users the opportunity to have the same behaviour as it used to be.

src/test/resources/logging.properties
```diff
@@ -0,0 +1,5 @@
+handlers = java.util.logging.ConsoleHandler
+java.util.logging.SimpleFormatter.format = [%1$tF %1$tT][%4$-6s][%2$s] %5$s %6$s %n
+java.util.logging.ConsoleHandler.level = ALL
+.level= INFO
+hudson.tasks.junit.level=ALL
```
jglick, line 5: Delete.

pom.xml
```diff
@@ -205,4 +211,20 @@
...
     </dependencyManagement>
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <configuration>
+                        <systemPropertyVariables>
+                            <java.util.logging.config.file>src/test/resources/logging.properties</java.util.logging.config.file>
```
jglick, line 222: You do not need to do this. Use https://jenkinsci.github.io/maven-hpi-plugin/run-mojo.html#loggers instead.

pom.xml
```diff
@@ -18,6 +18,7 @@
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <jenkins.version>2.332.1</jenkins.version>
         <no-test-jar>false</no-test-jar>
+        <slaveDebugPort>0</slaveDebugPort>
```
jglick, line 21:
```suggestion
```

pom.xml
```diff
@@ -205,4 +211,20 @@
...
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <configuration>
+                        <systemPropertyVariables>
+                            <java.util.logging.config.file>src/test/resources/logging.properties</java.util.logging.config.file>
+                            <org.jvnet.hudson.test.HudsonTestCase.slaveDebugPort>${slaveDebugPort}</org.jvnet.hudson.test.HudsonTestCase.slaveDebugPort>
```
jglick, line 223:
```suggestion
```

jglick:
Partly reverting #348?

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -48,9 +50,13 @@
 @Extension // see TestResultParser.all
 public class JUnitParser extends TestResultParser {
 
+    private static final Logger LOGGER = Logger.getLogger(JUnitParser.class.getName());
+
     private final boolean keepLongStdio;
     private final boolean allowEmptyResults;
 
+    private final boolean parseOldReports;
```
jglick, line 58: Although I normally advocate for options to be named in the positive rather than negative sense, in this case the behavior is sufficiently surprising that I would suggest it be called `skipOldReports` for example.

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -104,47 +115,66 @@ public TestResult parseResult(String testResultLocations, Run<?,?> build, FilePa
...
             if (files.length > 0) {
-                result = new TestResult(buildTime + (nowSlave - nowMaster), ds, keepLongStdio, pipelineTestDetails);
+                // not sure we can rely seriously on those timestamp so let's take the smaller one...
+                long filesTimestamp = Math.min(buildStartTimeInMillis,buildTimeInMillis);
+                // previous mode buildStartTimeInMillis + (nowSlave - nowMaster);
+                if (debug) {
+                    listener.getLogger().println("ParseResultCallable#invoke buildStartTimeInMillis:" + buildStartTimeInMillis
+                        + ",buildTimeInMillis:" + buildTimeInMillis + ",filesTimestamp:" + filesTimestamp + ",nowSlave:"
+                        + nowSlave + ",nowMaster:" + nowMaster);
+                    }
```
jglick, line 175: Seems like something better sent to `java.util.logging.Logger`?

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -178,51 +198,60 @@ public TestResult getTestResult() {
...
-        parse(buildTime, results, null);
+    public void parse(long filesTimestamp, DirectoryScanner results) throws IOException {
+        parse(filesTimestamp, results, null);
     }
 
     /**
      * Collect reports from the given {@link DirectoryScanner}, while
      * filtering out all files that were created before the given time.
-     * @param buildTime Build time.
+     * @param filesTimestamp file timestamp to filter files older.
```
jglick, line 208: Renaming variables like this makes it harder to review the actually changed behavior.

src/main/java/hudson/tasks/junit/TestResult.java
jglick: https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsRule.html#showAgentLogs-hudson.model.Slave-org.jvnet.hudson.test.LoggerRule- will probably do what you want.

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -195,6 +195,7 @@ public synchronized int getTotalCount() {
         }
         if(totalCount==null)
             getResult();    // this will compute the result
+
```
jglick, line 198: Please try to minimize the diff when filing a PR.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
jglick: FWIW my advice would be for the simpler behavior of #348 to be the default, and you would need to opt into this special workaround for misconfigured build tools.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-parseOldReports.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    If this option is unchecked, then the plugin will not parse report files older than the start of the build.
+    If this option is checked, then the build will parse and include in the report the test files older than the build start.
```
jglick, line 3:
```suggestion
    If this option is checked, then the build will parse and include in the report all test files matching the specified pattern, regardless of timestamp.
```

src/test/java/hudson/tasks/junit/CaseResultTest.java
```diff
@@ -222,42 +225,38 @@ public void testAge() throws Exception {
...
-        //First build result analysis:
-        TestResult tr = b1.getAction(TestResultAction.class).getResult();
-        assertEquals(1,tr.getFailedTests().size());
-        CaseResult cr = tr.getFailedTests().get(0);
-        assertEquals(1,cr.getAge()); //First execution, failing test age is expected to be 1
-
 
+        //First build execution
+        {
+            FreeStyleBuild b1 = rule.assertBuildStatus(Result.UNSTABLE, p.scheduleBuild2(0).get());
```
jglick, line 241: Is something being edited here? If so, it is hard to follow what. If you are adding test coverage for the new option, can it be kept to just creating a new test case method?

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -96,20 +98,24 @@
...
         project.getPublishersList().add(archiver);
-
         project.getBuildersList().add(new TouchBuilder());
     }
 
     @LocalData("All")
     @Test public void basic() throws Exception {
-        FreeStyleBuild build = project.scheduleBuild2(0).get(10, TimeUnit.SECONDS);
+        FreeStyleBuild build =
+                j.assertBuildStatus(Result.UNSTABLE, j.waitForCompletion(project.scheduleBuild2(0).waitForStart()));
```
jglick, line 118: Is this related somehow or just general cleanup that could be in another PR?

src/test/java/hudson/tasks/junit/TestResultTest.java
```diff
@@ -283,6 +286,46 @@ public void testTestcaseWithEmptyName() throws Exception {
         assertEquals("(?)", case1.getName());
     }
 
+    @Test
+    public void skipOldReports() throws Exception {
+        long start = System.currentTimeMillis();
+        File testResultFile1 = new File("src/test/resources/hudson/tasks/junit/old-reports/junit-report-1.xml");
+        Files.setLastModifiedTime(testResultFile1.toPath(), FileTime.fromMillis(start + 10));
+        File testResultFile2 = new File("src/test/resources/hudson/tasks/junit/old-reports/junit-report-2.xml");
+        Files.setLastModifiedTime(testResultFile2.toPath(), FileTime.fromMillis(start - 4000));
```
jglick, line 295: Beware that timestamp modifications on Windows are often quite imprecise.

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -257,6 +302,9 @@ public void stageInParallel() throws Exception {
 
         j.setDefinition(new CpsFlowDefinition("stage('outer') {\n" +
                 "  node {\n" +
+                "    touch 'first-result.xml'\n" +
+                "    touch 'second-result.xml'\n" +
+                "    touch 'third-result.xml'\n" +
```
jglick, line 307: Why is this sort of change necessary? If it was passing before #348, why would it fail now?

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -422,26 +475,16 @@ private void copyToWorkspace(WorkflowJob j, URL source, String destination) thro
...
     private static Predicate<FlowNode> stageForName(final String name) {
-        return new Predicate<FlowNode>() {
-            @Override
-            public boolean apply(@Nullable FlowNode input) {
-                return input instanceof StepStartNode &&
-                        ((StepStartNode) input).getDescriptor() instanceof StageStep.DescriptorImpl &&
-                        input.getDisplayName().equals(name);
-            }
-        };
+        return input -> input instanceof StepStartNode &&
```
jglick, line 485: Unrelated?

src/main/java/hudson/tasks/junit/TestResult.java
olamy: great I didn't know this option

pom.xml
olamy: when running test? 
As some tests were flaky I needed some way to have logs when running tests in https://ci.jenkins.io/

pom.xml
olamy: without those option how can I activate this when running tests from a console?

src/test/java/hudson/tasks/junit/CaseResultTest.java
olamy: read current version of the tests. there is some bad copy/paste which doesn't make sense.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
olamy: just removing a pattern which can be flaky

src/test/java/hudson/tasks/junit/TestResultTest.java
olamy: that's why I finally added this https://github.com/jenkinsci/junit-plugin/pull/384/files#diff-d80b11180304e55de1556e79fdbd96a795037aeca3c3817cc52aa5734e966b0eR122

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
olamy: in theory it is because files are copied to the workspace before the build start so without a touch they should be ignored by the parsing. but finally I added https://github.com/jenkinsci/junit-plugin/pull/384/files#diff-d80b11180304e55de1556e79fdbd96a795037aeca3c3817cc52aa5734e966b0eR122 which was something existing before #348 but hardcoded somewhere in the code https://github.com/jenkinsci/junit-plugin/pull/348/files#diff-d80b11180304e55de1556e79fdbd96a795037aeca3c3817cc52aa5734e966b0eL224

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
olamy: Accidentally click on IDE warning.

src/main/java/hudson/tasks/junit/JUnitParser.java
olamy: yes definitely not proud of this :)  but I didn't know the `#showAgentLogs` method. This will change my life for sure!
 I will change that.

src/main/java/hudson/tasks/junit/TestResult.java
olamy: frankly the previous name was just not really accurate

src/main/java/hudson/tasks/junit/JUnitParser.java
olamy: the name was according to the behaviour I changed back as with `false` as default.
if we choose `skipOldReports` (with `false` as default) so we do not return to the previous behaviour :) 
if there is some compromise for that no problem I will change that.
@jtnord ?

jtnord (approved):

jtnord:
should be noted 
Fixes: #368

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
jglick: Not sure I followed that. Maybe @timja did? https://github.com/jenkinsci/junit-plugin/pull/384/files#diff-d80b11180304e55de1556e79fdbd96a795037aeca3c3817cc52aa5734e966b0eR122 points to a blank line? Tip: if you link to blob URLs with hashes (not branch names) and line numbers/ranges, GH will render the code inline, which makes it a lot easier to follow discussions.

pom.xml
jglick: Oh sorry this is about test cases; the above is for `mvn hpi:run`. Use https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/LoggerRule.html

pom.xml
jglick: Is this for setting breakpoints in code run inside an agent JVM? If so, that should be automatic I think when you run Maven in the debugger. If that is not happening, then it is something that can be improved in `jenkins-test-harness`. I suspect you rediscovered https://github.com/jenkinsci/jenkins-test-harness/pull/428#discussion_r863031976? At any rate, do not commit stuff like this to plugin POMs. Either fix JTH, or temporarily edit your POM or whatever script you use to run tests in a debugger, but do not include the edit in commits (I have come to grumblingly accept `git add -p` as a necessity).

src/main/java/hudson/tasks/junit/TestResult.java
jglick: Probably so, but it would make your PR easier to understand and review if that were kept to a separate follow-up `chore` PR that claims to not change behavior.

pom.xml
olamy: ok I changed that too.
But it means this need to add some coding in all test classes which is a bit less convenient than changing a single file....

car-roll (approved):

pom.xml
olamy: using something such ` mvn test -Dtest=JUnitResultArchiverTest#slave -Dmaven.surefire.debug` doesn't start the slave in debug mode.
upgrading to last `1752.v86627a_c48d91` I went from 
```
$ "/Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/bin/java"  -Djava.awt.headless=true -Xmx512M -jar "/Volumes/workspace/dev/sources/jenkins/junit-plugin/target/jenkins-for-test/WEB-INF/lib/remoting-4.12.jar"
```

to 
```
$ "/Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/bin/java"  -Djava.awt.headless=true -Xmx512m -XX:+PrintCommandLineFlags -jar "/Volumes/workspace/dev/sources/jenkins/junit-plugin/target/jenkins-for-test/WEB-INF/lib/remoting-4.12.jar"
-XX:G1ConcRefinementThreads=13 -XX:GCDrainStackTargetSize=64 -XX:InitialHeapSize=536870912 -XX:MaxHeapSize=536870912 -XX:+PrintCommandLineFlags -XX:ReservedCodeCacheSize=251658240 -XX:+SegmentedCodeCache -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseG1GC 
```
not sure what is the need of those per default extra jvm setup? 

> At any rate, do not commit stuff like this to plugin POMs. 

why? what is the technical problem here? there is a default which do not activate anything.
using an option such if '-Dmaven.surefire.debug` then automatically run the slave in debug. why?
with such configuration it's possible to default what is happening on the unit but not on the slave or activate both or activate only debug in the slave.

jglick: > why?

Because debugging setup has nothing to do with your change. If such POM configuration is needed—which I doubt—it should be added to `plugin-pom`, not a particular plugin. If you need to temporarily `stash` some edits to help develop a change, fine, but do not commit them.

Most likely what is needed is a simple change to JTH to automatically activate the debugger on agent JVMs when the test case is being debugged, which is something that already happens for the controller implicitly in `JenkinsRule` but explicitly in `RealJenkinsRule`.

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
jglick: Example of the URL format, in case it helps: `https://github.com/jenkinsci/junit-plugin/blob/b232c5f48051edc49661726cbd03c49c0aee2a67/src/main/java/hudson/tasks/junit/TestResult.java#L223-L224` rendered as https://github.com/jenkinsci/junit-plugin/blob/b232c5f48051edc49661726cbd03c49c0aee2a67/src/main/java/hudson/tasks/junit/TestResult.java#L223-L224

pom.xml
olamy: > > why?
> 
> Because debugging setup has nothing to do with your change. If such POM configuration is needed—which I doubt—it should be added to `plugin-pom`, not a particular plugin. If you need to temporarily `stash` some edits to help develop a change, fine, but do not commit them.

I removed because it looks forbidden even if there is no other way to do it...


> 
> Most likely what is needed is a simple change to JTH to automatically activate the debugger on agent JVMs when the test case is being debugged, which is something that already happens for the controller implicitly in `JenkinsRule` but explicitly in `RealJenkinsRule`.

olamy:
review has been done, changes has been made according to the comments.
What else? what's next? merging and releasing maybe to move this forward?
thanks

timja:
I have not had time to properly evaluate the impact of not applying https://github.com/jenkinsci/junit-plugin/pull/384#discussion_r874746884

I have not had time to test it interactively either.

----

If another maintainer wishes to do the above then feel free to proceed and not wait for me

olamy:
@timja we are talking here of restoring a feature/behaviour which was here for ages... and without remplacement or even a configuration mode to activate or de activate it.
This PR simply restore a backward compat and even make this configurable.
what is the problem?

timja (changes requested):
This breaks the gradle setup in a similar way too before

Please re-evaluate https://github.com/jenkinsci/junit-plugin/pull/384#discussion_r874746884

Pipeline I used (same as the one used in the last bug):

To reproduce run it twice and you get:
> ERROR: None of the test reports contained any result

```
node {
    stage('checkout') {
      git "https://github.com/hmcts/sds-toffee-recipes-service"
    }
    
    stage('Build') {
        sh './gradlew build'
    }
    
    stage('Test') {
        try {
          sh './gradlew test'
        } finally {
            junit 'build/test-results/**/TEST-*.xml'
        }
    }
    stage('Test2') {
        try {
          sh './gradlew test'
        } finally {
            junit 'build/test-results/**/TEST-*.xml'
        }
    }
}
```

olamy:
> This breaks the gradle setup in a similar way too before
> 
> Please re-evaluate [#384 (comment)](https://github.com/jenkinsci/junit-plugin/pull/384#discussion_r874746884)
> 
> Pipeline I used (same as the one used in the last bug):
> 
> To reproduce run it twice and you get:
> 
> > ERROR: None of the test reports contained any result
> 
> ```
> node {
>     stage('checkout') {
>       git "https://github.com/hmcts/sds-toffee-recipes-service"
>     }
>     
>     stage('Build') {
>         sh './gradlew build'
>     }
>     
>     stage('Test') {
>         try {
>           sh './gradlew test'
>         } finally {
>             junit 'build/test-results/**/TEST-*.xml'
>         }
>     }
>     stage('Test2') {
>         try {
>           sh './gradlew test'
>         } finally {
>             junit 'build/test-results/**/TEST-*.xml'
>         }
>     }
> }
> ```

use
 
```
junit 'build/test-results/**/TEST-*.xml' parseOldReports: true
```
Remember the goal of this change is to restore backward compat which has been broken and now it is configurable.

olamy:
so as it looks nobody want to move forward and it looks to be possible now in Jenkins to break backward compat and calling this fixing a bug....
So this PR restore backward compat with the new backward compat (well I'm lost now if it's restoring something or breaking something).

olamy:
> > So finally this PR restore backward compat with the new backward compat (well I'm lost now if it's restoring something or breaking something).
> 
> it re-breaks it with an option in place to opt-out.
well to be precise it breaks a previous breaking change...
> 
> If you swapped the option around (as suggested by Jesse) I would be completely fine with it.
>

last commit does it.

 
> > use
> > junit 'build/test-results/**/TEST-*.xml' parseOldReports: true
> 
> Aware, I don't think that should be the default making people set that.
> 
> > so as it looks nobody want to move forward and it looks to be possible now in Jenkins to break backward compat and calling this fixing a bug....
> 
> It was an intentional behaviour change.

it's marked as bug fix in the release note and not as a breaking change.  and frankly there was no error.

Screen Shot 2022-05-26 at 7 14 48 pm



> 
> > And frankly I have better waves to surf (that's the Australian version of the better fish...)
> 
> :D sure, no objections to the PR in principle, but just need the option adjusted imo.

timja (approved):

timja:
@jglick did you want to re-review?

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -107,7 +107,7 @@ public class JUnitResultArchiver extends Recorder implements SimpleBuildStep, JU
      */
     private boolean skipMarkingBuildUnstable;
 
-    private boolean parseOldReports;
+    private boolean parseOldReports = true;
```
jglick, line 110: Beware that this will not take effect when loading existing non-Pipeline jobs from disk: the field will be missing from `config.xml` and so will default to `false`. Meaning the pre-#348 behavior. Whereas for Pipeline jobs the `JUnitResultStep` object is created via constructor every time, so it will default to the #348 behavior. Not necessarily a problem, but possibly a subtle point of confusion. If you want to switch to the #348 behavior by default when loading existing traditional jobs, you could either
* invert the sense of the flag, so default ~ `false` is the #348-style behavior, as in https://github.com/jenkinsci/junit-plugin/pull/384#discussion_r874738335
* convert to `Boolean` and use `readResolve` to translate `null` to `true`

pom.xml
```diff
@@ -187,6 +187,11 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jackson2-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-utility-steps</artifactId>
+            <scope>test</scope>
```
jglick, line 193: Is this just for the `touch` step? For a cross-OS replacement, try

```groovy
writeFile file: 'x.xml', text: readFile('x.xml')
```

I guess so long as this plugin is in the BOM it is not a burden to use it in tests, though.

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -63,18 +69,23 @@ public JUnitParser() {
      */
     @Deprecated
     public JUnitParser(boolean keepLongStdio) {
-        this.keepLongStdio = keepLongStdio;
-        this.allowEmptyResults = false;
+        this(keepLongStdio , false, false);
```
jglick, line 72:
```suggestion
        this(keepLongStdio , false, true);
```

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -63,18 +69,23 @@ public JUnitParser() {
...
     }
 
     /**
      * @param keepLongStdio if true, retain a suite's complete stdout/stderr even if this is huge and the suite passed
      * @param allowEmptyResults if true, empty results are allowed
      * @since 1.10
      */
+    @Deprecated
     public JUnitParser(boolean keepLongStdio, boolean allowEmptyResults) {
+        this(keepLongStdio , allowEmptyResults, false);
```
jglick, line 82:
```suggestion
        this(keepLongStdio , allowEmptyResults, true);
```

src/main/java/hudson/tasks/junit/JUnitParser.java
```diff
@@ -104,47 +115,66 @@ public TestResult parseResult(String testResultLocations, Run<?,?> build, FilePa
...
+
+        private final long buildTimeInMillis;
         private final String testResults;
         private final long nowMaster;
         private final boolean keepLongStdio;
         private final boolean allowEmptyResults;
         private final PipelineTestDetails pipelineTestDetails;
         private final TaskListener listener;
 
+        private boolean parseOldReports = true;
```
jglick, line 143:
```suggestion
        private final boolean parseOldReports;
```

src/main/java/hudson/tasks/junit/JUnitTask.java
```diff
@@ -16,4 +16,7 @@ public interface JUnitTask {
     boolean isSkipPublishingChecks();
 
     String getChecksName();
+
+    boolean isParseOldReports();
```
jglick, line 20:
```suggestion
    default boolean isParseOldReports() {
       return true;
    }
```
https://github.com/jenkinsci/uipath-automation-package-plugin/blob/b391e63b45a924782c36b12de4047641e3d910c8/src/main/java/com/uipath/uipathpackage/UiPathTest.java#L44 whatever that is

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -129,27 +141,37 @@ public TestResult(boolean keepLongStdio) {
...
+    public TestResult(long filesTimestamp, DirectoryScanner results, boolean keepLongStdio) throws IOException {
+        this(filesTimestamp, results, keepLongStdio, null);
     }
 
+    @Deprecated
+    public TestResult(long filesTimestamp, DirectoryScanner results, boolean keepLongStdio,
+                      PipelineTestDetails pipelineTestDetails) throws IOException {
+        this.keepLongStdio = keepLongStdio;
+        impl = null;
+        parse(filesTimestamp, results, pipelineTestDetails);
```
jglick, line 158: Should probably be made into a `this`-constructor call and `parseOldReports` made `final` here?

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -106,13 +113,18 @@ public final class TestResult extends MetaTabulatedResult {
 
     private float duration;
 
+    private boolean parseOldReports = true;
```
jglick, line 116: Again will not be read from disk as such.

Could this field be `transient`? I imagine `parse` is only called from the same Jenkins session as the `TestResult` constructor? I think it does need to be a field rather than a local variable, unfortunately, due to https://github.com/jenkinsci/junit-plugin/blob/83cdce2b26ab579da34b344063b599bfe7985c6f/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java#L256-L258 if I read it right.

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -178,51 +200,62 @@ public TestResult getTestResult() {
...
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    LOGGER.fine("file " + reportFile + " not parsed: parseOldReports-" + parseOldReports
+                            + ",lastModified:" + Files.getLastModifiedTime(reportFile.toPath()).toMillis() + ",filesTimestamp:" + filesTimestamp);
+                }
+                continue;
+            }
             // only count files that were actually updated during this build
             parsePossiblyEmpty(reportFile, pipelineTestDetails);
         }
+        if (LOGGER.isLoggable(Level.FINE)) LOGGER.fine("testSuites size:" + this.getSuites().size());
```
jglick, line 258:
```suggestion
        if (LOGGER.isLoggable(Level.FINE)) LOGGER.fine(() -> "testSuites size:" + this.getSuites().size());
```

generally

src/main/java/hudson/tasks/junit/TestResult.java
```diff
@@ -934,4 +974,5 @@ public CaseResult getCase(String suiteName, String transformedFullDisplayName) {
                 .findFirst()
                 .orElse(null);
     }
+
```
jglick, line 977:
```suggestion
```

src/main/java/hudson/tasks/test/DefaultTestResultParserImpl.java
```diff
@@ -96,7 +96,7 @@ public TestResult invoke(File dir, VirtualChannel channel) throws IOException, I
                 List<File> files = new ArrayList<>(paths.length);
                 for (FilePath path : paths) {
                     File report = new File(path.getRemote());
-                    if (ignoreTimestampCheck || localBuildTime - 3000 /*error margin*/ < report.lastModified()) {
+                    if (ignoreTimestampCheck || localBuildTime - hudson.tasks.junit.TestResult.FILE_TIME_PRECISION_MARGIN < report.lastModified()) {
                         // this file is created during this build
```
jglick, line 100: I guess this was forgotten in #348?

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -56,4 +56,7 @@ THE SOFTWARE.
     <f:entry title="${%Skip marking build as unstable on test failure}" field="skipMarkingBuildUnstable">
             <f:checkbox default="false" title="${%If checked, the test failures will still be reported but won't mark the build as unstable}"/>
     </f:entry>
+    <f:entry title="${%Include report files older than build start}" field="parseOldReports">
+            <f:checkbox default="true" title="${%If checked, the test report files older than build start will be included in the parsing}"/>
```
jglick, line 60:
```suggestion
            <f:checkbox default="true" title="${%If unchecked, test report files older than the build start time will be omitted}"/>
```

src/test/java/hudson/tasks/junit/CaseResultTest.java
```diff
@@ -222,42 +225,38 @@ public void testAge() throws Exception {
...
-        //First build execution:
+        // add builder copying test result file
         p.getBuildersList().add(new TestBuilder() {
             public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-                build.getWorkspace().child("junit.xml").copyFrom(
-                    getClass().getResource(testResultResourceFile));
+                FilePath junitFile = build.getWorkspace().child("junit.xml");
+                junitFile.copyFrom(getClass().getResource(testResultResourceFile));
+                // sadly this can be flaky for 1ms.... (but hey changing to nano even in core might complicated :))
+                junitFile.touch(System.currentTimeMillis()+1L);
```
jglick, line 234: So if `parseOldReports` now defaults to true, all the complicated logic to test this feature can be isolated to a handful of test cases, leaving the rest of the test code free of flaky timestamp manipulation, right?

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -96,20 +100,27 @@
...
```
jglick, line 123:
```suggestion
        FreeStyleBuild build = j.buildAndAssertStatus(Result.UNSTABLE, project);
```

As in https://github.com/jenkinsci/junit-plugin/pull/384#discussion_r874749729 it would be better to file such cleanups as a separate `chore` / `tests` PR that explicitly makes no change to behavior and could be merged earlier (or later) to avoid confusing and delaying the review of this PR.

src/test/java/hudson/tasks/junit/JUnitResultArchiverTest.java
```diff
@@ -225,8 +237,9 @@ private HtmlForm findForm(HtmlPage page, String action) {
     }
     private void doRepeatedArchiving(boolean slave) throws Exception {
         if (slave) {
-            DumbSlave s = j.createOnlineSlave();
-            project.setAssignedLabel(s.getSelfLabel());
+            DumbSlave s = j.createSlave("label1 label2", null);
+            project.setAssignedLabel(j.jenkins.getLabel("label1"));
```
jglick, line 241: Any particular reason?

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -348,7 +398,7 @@ public void currentBuildResultUnstable() throws Exception {
         WorkflowJob j = rule.jenkins.createProject(WorkflowJob.class, "currentBuildResultUnstable");
         j.setDefinition(new CpsFlowDefinition("stage('first') {\n" +
                 "  node {\n" +
-                "    def results = junit(testResults: '*.xml')\n" + // node id 7
+                "    def results = junit(testResults: '*.xml', parseOldReports: true)\n" + // node id 7
```
jglick, line 401:
```suggestion
                "    def results = junit(testResults: '*.xml')\n" + // node id 7
```

src/test/java/hudson/tasks/junit/pipeline/JUnitResultsStepTest.java
```diff
@@ -422,26 +475,16 @@ private void copyToWorkspace(WorkflowJob j, URL source, String destination) thro
...
-                return input != null &&
-                        input.getAction(LabelAction.class) != null &&
-                        input.getAction(ThreadNameAction.class) != null &&
-                        name.equals(input.getAction(ThreadNameAction.class).getThreadName());
-            }
-        };
+        return input -> input != null &&
+                input.getAction(LabelAction.class) != null &&
+                input.getAction(ThreadNameAction.class) != null &&
+                name.equals(input.getAction(ThreadNameAction.class).getThreadName());
```
jglick, line 481: Fine but again better left to a separate `chore` PR.

src/test/java/hudson/tasks/test/TestResultExtensionTest.java
```diff
@@ -62,8 +62,7 @@ public void testTrivialRecorder() throws Exception {
         assertNotNull("parent action should have an owner", action.run);
         Object resultObject = action.getResult();
         assertNotNull("we should have a result");
-        assertTrue("result should be an TestResult",
```
jglick, line 65: No apparent reason why this file should have been edited here at all.

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -79,7 +80,7 @@ public void extractChecksDetailsFailingSingleTestResult() throws Exception {
         WorkflowJob j = rule.jenkins.createProject(WorkflowJob.class, "singleStep");
         j.setDefinition(new CpsFlowDefinition("stage('first') {\n" +
                 "  node {\n" +
-                "    def results = junit(testResults: '*.xml')\n" +
+                "    def results = junit(testResults: '*.xml', parseOldReports: true)\n" +
```
jglick, line 83:
```suggestion
                "    def results = junit(testResults: '*.xml')\n" +
```

src/test/java/io/jenkins/plugins/junit/checks/JUnitChecksPublisherTest.java
```diff
@@ -106,7 +107,7 @@ public void extractChecksDetailsFailingMultipleTests() throws Exception {
...
```
jglick, line 110:
```suggestion
                "    def results = junit(testResults: '*.xml')\n" +
```

jglick (commented):
No blockers, though this is a big diff that could probably be cleaned up.

jtnord:
will need corresponding fixes in the `maven` plugin where the old behaviour is expected.

pom.xml
olamy: `touch` is cross-OS

olamy:
well as you don't want to restore backward compatibility per default. I have changed the configuration name to `skipOldReports` to better match of this non backward compatibility option with false per default as this will avoid so much over complicated and ugly code.

olamy:
> 
> ```
> node {
>     stage('checkout') {
>       git "https://github.com/hmcts/sds-toffee-recipes-service"
>     }
>     
>     stage('Build') {
>         sh './gradlew build'
>     }
>     
>     stage('Test') {
>         try {
>           sh './gradlew test'
>         } finally {
>             junit 'build/test-results/**/TEST-*.xml'
>         }
>     }
>     stage('Test2') {
>         try {
>           sh './gradlew test'
>         } finally {
>             junit 'build/test-results/**/TEST-*.xml'
>         }
>     }
> }
> ```

this is interesting. do you mean test results file from first stage are not parsed in 2nd step?
as the parsing is based on the start of the build those files should not ignored.
I'm interested of the timestamp of those files, Can run the same but add:
```
sh "ls -lrt 'build/test-results/**/TEST-*.xml'"
junit 'build/test-results/**/TEST-*.xml'
```

src/test/java/hudson/tasks/test/TestResultExtensionTest.java
olamy: correct I will restore the previous formatting to help people who have a maximum of 65 characters per line

src/test/java/hudson/tasks/test/TestResultExtensionTest.java
```diff
@@ -76,7 +76,6 @@ public void testTrivialRecorder() throws Exception {
         HtmlPage testReportPage = wc.getPage(project, "/lastBuild/testReport/");
         j.assertGoodStatus(testReportPage);
 
-
```
olamy, line 79: TO RESTORE!

timja (approved):

PR #435, opened 2022-09-23
Check the card and not the card-body in TestResultPublishingTest

raul-arabaolaza:
With `bootstrap5-api#5.2.0-3` the `history` text is inside a div with a `card-header` class instead inside the div with `card-body` this change will make the test work for both versions and IMO makes the test a bit less fragile, for example when running the PCT.

Updating the dependency would need a core bump and I am not sure how difficult that would be, and TBH I do not have time to do that. Anyway this change should be fine even after the core and boostrap5 update.


timja (approved):
if it works

timja:
I assume you want a release?

PR #439, opened 2022-10-01
[JENKINS-69658] Move internal `script` block in "failed-test.jelly" to external file

Jagrutiti:
- [X] Link to relevant issues in GitHub or Jira

## What
Moved the following the `<script>` block that was causing the issue to an external file:

```
 <script type="text/javascript">
      <!-- TODO make sure load doesn't happen every time -->
      function showFailureSummary(id,query) {
        var element = document.getElementById(id)
        element.style.display = "";
        document.getElementById(id + "-showlink").style.display = "none";
        document.getElementById(id + "-hidelink").style.display = "";

        if (typeof query !== 'undefined') {
          var rqo = new XMLHttpRequest();
          rqo.open('GET', query, true);
          rqo.onreadystatechange = function() { element.innerHTML = rqo.responseText; }
          rqo.send(null);
        }
      }

      function hideFailureSummary(id) {
        document.getElementById(id).style.display = "none";
        document.getElementById(id + "-showlink").style.display = "";
        document.getElementById(id + "-hidelink").style.display = "none";
      }
  </script>
```

## Closes
https://issues.jenkins.io/browse/JENKINS-69658


src/main/resources/lib/hudson/test/failed-test.jelly
```diff
@@ -35,28 +35,7 @@ THE SOFTWARE.
...
-        }
-      }
-
-      function hideFailureSummary(id) {
-        document.getElementById(id).style.display = "none";
-        document.getElementById(id + "-showlink").style.display = "";
-        document.getElementById(id + "-hidelink").style.display = "none";
-      }
-    </script>
+    <st:adjunct includes="hudson.test.js.failureSummary.js" />
```
Jagrutiti, line 38: I am not familiar with Jelly and its syntax.

uhafner: I think you need to use the path suffix that starts right after the folder `src/main/resources`. So `"lib.hudson.test.failureSummary"`. 

(Maybe it would be also good to rename the file to `failure-summary.js`, I'm not sure if the capitalization for files works all over the framework)

timja (commented):
the test failures show this isn't working

Jagrutiti:
The error says: `ReferenceError: "showFailureSummary" is not defined.`

There is still a path issue. I will read about Jelly, add a path in Jelly, and try again.

Jagrutiti:
Hey, @uhafner @timja 

This tutorial here:
https://wiki.eclipse.org/Understanding_Jelly_Tags
https://wiki.jenkins.io/display/JENKINS/Basic+guide+to+Jelly+usage+in+Jenkins

Do not explain `st: adjunct`; I cannot find other tutorials explaining the tag. And how to import files in Jelly.

 I did find the tutorial explaining `adjunct`: https://www.jenkins.io/doc/developer/views/exposing-bundled-resources/#adjuncts
 
I am not able to figure out where my file path is wrong. Or is it something else?

timja:
> I am not able to figure out where my file path is wrong. Or is it something else?

best to search the jenkinsci GitHub org to see how it's used, lots of this stuff is under-documented so code is the best example

Jagrutiti:
All the checks have passed. 

@uhafner was right about this:

>I think you need to use the path suffix that starts right after the folder src/main/resources

I was making the mistake of adding the `.js` extension to the end of the file.

uhafner (approved):
Thanks for your PR. The 1:1 conversions looks good. 

(I wonder if we should also change `var` to `const` and the TODO does not make sense to me as well?)

Jagrutiti:
About changing `var` to `const` t

The properties are being reassigned in two places.
```
function showFailureSummary(id,query) {
        var element = document.getElementById(id)
        element.style.display = "";                           // property changes here
        document.getElementById(id + "-showlink").style.display = "none";
        document.getElementById(id + "-hidelink").style.display = "";

        if (typeof query !== 'undefined') {
          var rqo = new XMLHttpRequest();
          rqo.open('GET', query, true);
          rqo.onreadystatechange = function() { element.innerHTML = rqo.responseText; } // property changes here
          rqo.send(null);
        }
      }
```

In that case should we change it to `const`?

timja:
> The properties are being reassigned in two places.

That's not reassigning the value that's modifying properties which is allowed when it's a `const`

src/main/resources/lib/hudson/test/js/failureSummary.js
```diff
@@ -0,0 +1,21 @@
+// TODO make sure load doesn't happen every time
```
timja, line 1: not sure what it's referring to but all seems fine here
```suggestion
```

src/main/resources/lib/hudson/test/js/failureSummary.js
```diff
@@ -0,0 +1,21 @@
...
+        rqo.onreadystatechange = function() { element.innerHTML = rqo.responseText; }
+        rqo.send(null);
+    }
+}
+
+function hideFailureSummary(id) {
+    document.getElementById(id).style.display = "none";
+    document.getElementById(id + "-showlink").style.display = "";
+    document.getElementById(id + "-hidelink").style.display = "none";
+}
```
timja, line 21:
```suggestion
}

```

timja (approved):
Thanks!

PR #444, opened 2022-10-07
[JENKINS-69656] Un-inlining AbstractTestResultAction/summary.jelly for CSP compliance

aneveux:
Hello there :wave:

Hacktoberfest PR here to solve [JENKINS-69656](https://issues.jenkins.io/browse/JENKINS-69656), by un-inlining AbstractTestResultAction/summary.jelly, and then making a step forward towards CSP adoption in Jenkins.

I tested that PR by following the details shared in the Jira ticket, and ensured my contribution didn't break the
current behavior.

While extracting the inline script to a js file, I also did a little bit of refactoring to the JavaScript source code, I
hope that's OK. Let me know if you don't like that part and I'll revert to the previous content of the JavaScript part.

Thanks a lot for your review, let me know if there's anything missing I can take care of :smile:

Have a great day!

### DIFF
--- a/src/main/resources/hudson/tasks/test/AbstractTestResultAction/show-failures.js
+++ b/src/main/resources/hudson/tasks/test/AbstractTestResultAction/show-failures.js
@@ -1,0 +1,33 @@
+/**
+* Displays all the tests failures, and hides the link allowing to display them from UI.
+*/
+function showFailures() {
+    // Displaying all the hidden elements from the page (those are the failed tests)
+    let hiddenElements = document.getElementsByClassName("hidden");
+
+    // DEV MEMO:
+    // hiddenElements is not an array but an HTMLCollection.
+    // To allow using forEach, we need an array, so I'm using the spread operator below to get that.
+    [...hiddenElements].forEach(element => { element.style.display = ""; });
+
+    // Now hiding the link from UI allowing to show all failed tests
+    let showFailuresLink = document.getElementById("showLink");
+    showFailuresLink.style.display = "none";
+}
+
+// Adding an onclick listener to the link in UI allowing to display all failed tests
+// DEV MEMO:
+// We are doing it after DOM content is loaded as a good practice to ensure we are not slowing down
+// the page rendering. In that particular situation the addition of the onclick handler shouldn't
+// really impact the page performances, but rather stick with good practices.
+
+document.addEventListener('DOMContentLoaded', (event) => {
+
+  // Retrieving the link from UI allowing to show all failed tests
+  // Note: we are retrieving the link by its ID to match how it was already done
+  // in the showFailures method above.
+  const showFailuresLink = document.getElementById("showLink");
+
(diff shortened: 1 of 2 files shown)


src/main/resources/hudson/tasks/test/AbstractTestResultAction/show-failures.js
```diff
@@ -0,0 +1,33 @@
...
+// We are doing it after DOM content is loaded as a good practice to ensure we are not slowing down
+// the page rendering. In that particular situation the addition of the onclick handler shouldn't
+// really impact the page performances, but rather stick with good practices.
+
+document.addEventListener('DOMContentLoaded', (event) => {
+
+  // Retrieving the link from UI allowing to show all failed tests
+  // Note: we are retrieving the link by its ID to match how it was already done
+  // in the showFailures method above.
+  const showFailuresLink = document.getElementById("showLink");
```
aneveux, line 29: Note:

The Jira issue ([JENKINS-69656](https://issues.jenkins.io/browse/JENKINS-69656)) mentioned:

> Be careful, the id "showLink" is perhaps too generic (could be present on other elements in the same page. Adding a className like "showFailuresLink" and using it as selector could be a good idea.

But the original piece of code was actually using `document.getElementById("showLink")` to identify the link, so I considered it safe to rely on the very same identification for adding the `onclick` handler.

If you feel like it's not good enough, I'm happy to change it, but the current proposal matches the original behavior of the source code.

Wadeck (approved):
**Advice:** do not change code when not necessary for such "move". That makes the review more complicated and could introduce "new" issues.

Adding "arrow functions", "let", "spread operator", could also add compatibility issue. I quickly checked, it seems OK with https://www.jenkins.io/doc/administration/requirements/web-browsers/.

Not manually tested with different browsers. OK in terms of moving code for CSP PoV.

src/main/resources/hudson/tasks/test/AbstractTestResultAction/show-failures.js
Wadeck: Perhaps a matter of luck, but we have several occurrences of `showLink` being used as id: https://github.com/search?l=XML&q=org%3Ajenkinsci+showLink&type=Code

aneveux: Do you want me to change it in that PR with something unique (both in the `onclick` and in the `showFailures` function)? 

Or should I create another issue to fix it? (or ignore it?)

aneveux:
:+1: Thanks for the advice! I was not sure about changing moved code actually, so I'll be careful not doing it in future PRs.

As per browser compability, I assumed it was OK using things which were there for years, and checked that it was in line with Level 2 of Jenkins web-browsers compatibility (as Level 3 is basically everything).

I manually tested with Brave, Chromium, Firefox on Linux.

timja (approved):

Issue #494, opened 2023-02-17
Make it possible to include custom columns from TestActions in the ClassResult table


jonaslind:
### What feature do you want to see added?

The `TestDataPublisher` extension in junit-plugin allows other plugins to contribute `TestAction`s that enrich test result UI. This is called "Additional test report features" in the job configuration UI. There are currently jelly hooks that allow other plugins to provide custom content in various places across the test result pages.

I suggest that two new optional jelly hooks are added that allow `TestAction`s to add custom columns to the `ClassResult` table: one hook for the table headers and one hook for the table rows. That way, it would be possible to include additional column headers and row cells from `TestAction`s.

Custom columns in the `ClassResult` table can be useful for several things. For example, [JENKINS-12575](https://issues.jenkins.io/browse/JENKINS-12575) discusses chronological ordering of test cases. The de-facto JUnit XML report schema as implemented by various software including the junit-plugin doesn't support datetime information in test cases, so a generic catch-all support for chronological ordering may not be a good fit for the standard junit-plugin itself. Instead, the junit-plugin by default sorts test cases alphabetically by name, and the UI allows the user to re-sort the table by the other standard columns: duration and status.

However, there may be specific cases where a more specialized third party plugin can deduce datetime information. If such a plugin can add a custom column (for example "Timestamp") to the `ClassResult` table, the user can choose to sort the table chronologically. I've attached a screenshot of a mockup showing the `ClassResult` table sorted by the values in the Timestamp column.

Note that the Timestamp column is just an example of a custom column. There is nothing in this suggested improvement that prevents other custom columns to be added by other plugins. There is also nothing in this suggested improvement specific to timestamps.

In more concrete terms, I am suggesting that two new optional `classresult-tableheader.jelly` and `classresult-tablerow.jelly` pages are added to `TestAction`s and included in the `hudson/tasks/junit/ClassResult/body.jelly` resource similar to how the `badge.jelly` page is included today.

```xml
diff --git a/src/main/resources/hudson/tasks/junit/ClassResult/body.jelly b/src/main/resources/hudson/tasks/junit/ClassResult/body.jelly
index 6ce0000..2bb9e88 100644
--- a/src/main/resources/hudson/tasks/junit/ClassResult/body.jelly
+++ b/src/main/resources/hudson/tasks/junit/ClassResult/body.jelly
@@ -32,6 +32,9 @@ THE SOFTWARE.
           <th class="pane-header">${%Test name}</th>
[...]

PR #503, opened 2023-03-08
Make it possible to include custom columns from TestActions in case, class and package tables

jonaslind:
Add optional jelly hooks that allow TestActions to contribute custom columns to tables that list CaseResults, ClassResults and PackageResults.

More details in the corresponding enhancement issue:
Fixes https://github.com/jenkinsci/junit-plugin/issues/494

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestAction.java
+++ b/src/main/java/hudson/tasks/junit/TestAction.java
@@ -33,6 +33,10 @@ import org.kohsuke.stapler.export.ExportedBean;
  * <li>index.jelly: included at the top of the test page</li>
  * <li>summary.jelly: included in a collapsed panel on the test parent page</li>
  * <li>badge.jelly: shown after the test link on the test parent page</li>
+ * <li>casetableheader.jelly: allows additional table headers to be shown in tables that list test methods</li>
+ * <li>classtableheader.jelly: allows additional table headers to be shown in tables that list test classes</li>
+ * <li>packagetableheader.jelly: allows additional table headers to be shown in tables that list test packages</li>
+ * <li>tablerow.jelly: allows additional table cells to be shown in tables that list test methods, classes and packages</li>
  * </ul>
  * 
  * @author tom
--- a/src/main/java/hudson/tasks/test/TabulatedResult.java
+++ b/src/main/java/hudson/tasks/test/TabulatedResult.java
@@ -121,4 +121,14 @@ public abstract class TabulatedResult extends TestResult {
     public String getChildTitle() {
         return "";
     }
+
+    /**
+     * Get a simple name for the type of children the {@link #getChildren()} method returns, for example "case", "class"
+     * or "package".
+     *
+     * @return the type of children this result has, all lowercase.
+     */
+    public String getChildType() {
+        return "";
+    }
 }
--- a/src/main/java/hudson/tasks/junit/ClassResult.java
+++ b/src/main/java/hudson/tasks/junit/ClassResult.java
@@ -102,6 +102,11 @@ public final class ClassResult extends TabulatedResult implements Comparable<Cla
         return "Class Results";
     }
 
+    @Override
+    public String getChildType() {
+        return "case";
+    }
+
     @Exported(visibility=999)
     @Override
     public String getName() {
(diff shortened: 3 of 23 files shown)


jonaslind:
Desired reviewer: @timja

timja:
can you check the build is green before requesting my review please

jonesbusy:
Looks also the master branch is broken : https://ci.jenkins.io/job/Plugins/job/junit-plugin/job/master/321/

jonaslind:
@timja @jonesbusy thanks for comments!

Yes, the reason the build fails is RequireUpperBoundDeps, and this issue comes from the master branch. It's not introduced by my pull request. For example, I saw the same failure in @timja 's pull request #501 that was merged yesterday:

https://ci.jenkins.io/job/Plugins/job/junit-plugin/view/change-requests/job/PR-501/1/

It looks like the issue has been introduced by dependabot since c789200cf343cef9d806e394f050eac104c700c0. At the same time, it looks like @timja is replacing dependabot with renovate?

I'm happy to try and tackle the RequireUpperBoundDeps failure if no-one gets to it before me. Am I expected to fix that in this pull request, or shall it be done separately in a different pull request?

timja:
Weird given CI passed on that build.

I've switched it to renovate but they are basically the same tools with a slightly different feature set.

If you're happy to do it then a separate PR would be great. Otherwise I'll get to it when I have a chance

jonesbusy:
@jonaslind I'm not maintainer of this plugin, but same things happen to me on the skip-notifications-trait-plugin 

https://github.com/jenkinsci/skip-notifications-trait-plugin/pull/62/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R59

jonaslind:
@timja , do you think you'll be able to review this pull request?

timja:
At some point yes but I don't have time right now, I've requested review from all the maintainers though

src/main/java/hudson/tasks/junit/TestAction.java
```diff
@@ -33,6 +33,8 @@
  * <li>index.jelly: included at the top of the test page</li>
  * <li>summary.jelly: included in a collapsed panel on the test parent page</li>
  * <li>badge.jelly: shown after the test link on the test parent page</li>
+ * <li>classresult-tableheader.jelly: allows additional table headers to be shown in the test class' table of test methods</li>
+ * <li>classresult-tablerow.jelly: allows additional table cells to be shown in the test class' table of test methods</li>
```
timja, line 37:
```suggestion
 * <li>classresult-tableheader.jelly: allows additional table headers to be shown in the test class's table of test methods</li>
 * <li>classresult-tablerow.jelly: allows additional table cells to be shown in the test class's table of test methods</li>
```

timja (approved):
Seems fine, is there a reason you're just doing this for class results and not other ones?

src/main/java/hudson/tasks/junit/TestAction.java
jonaslind: Thank you for the grammar fix, will include in next commit!

timja:
I think they are as you click through test rests from the class level to the individual test level.

Its often hard to tell where they are used, I generally put some gibberish in each file and click around till I trace them back

jonaslind:
@timja, here's a new commit with custom columns added to `MetaTabulatedResult`. Turns out that `History` is not a `TestObject`, so adding custom columns to history pages would involve more refactoring. I'm not sure that is motivated unless there is a clear use case for it. So I've skipped that for now.

Please have a look when you have the time. Thanks!

src/main/java/hudson/tasks/test/TabulatedResult.java
```diff
@@ -121,4 +121,8 @@ public TabulatedResult blockToTestResult(@NonNull PipelineBlockWithTests block,
     public String getChildTitle() {
         return "";
     }
+
+    public String getChildType() {
```
timja, line 125: please add javadoc

timja (commented):
looks good, just one docs change missing

timja (approved):
Thanks!

jonaslind:
Thank you very much @timja for all your help and input!

PR #520, opened 2023-04-24
[JENKINS-71139] Alternate reproducer for NUL issue in `SuiteResult.stdout`

jglick:
https://github.com/jenkinsci/jenkins/pull/7875 but demonstrating how this can break JUnit test results when a test happens to print NUL characters in its standard output, depending on exactly how the report is structured. Passes against Jenkins 2.397, fails against 2.398—though, weirdly, in a different way than both the observed file load error and the core test failure: rather than the parsing failing with an error message, the `stdout` field is simply quietly omitted (left `null`).


src/test/resources/hudson/tasks/junit/null-output.txt
jglick, line 1:
```console
$ od -a src/test/resources/hudson/tasks/junit/null-output.txt
0000000   f   o   o nul   b   a   r
0000007
```

PR #541, opened 2023-06-20
Add support for 'properties'

jjhiblot:
This PR adds support for the 'properties'  in a junit file.
More specifically it parses the properties node in the JUnit file and the properties of the the testcases are displayed along the other elements of the test execution (stdout, stderr, ...) 
Properties are often used to give describe the context of a test execution and having this information under the eye when looking at a test result seems valuable.

Properties are also used to attach files to a test result. This has not been handled in this PR.

This work is based on the ideas of Thomas Dee (https://github.com/jenkinsci/junit-plugin/pull/129)

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -46,6 +46,8 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -78,6 +80,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     private final String skippedMessage;
     private final String errorStackTrace;
     private final String errorDetails;
+    private final Map<String, String> properties;
     @SuppressFBWarnings(value = "SE_TRANSIENT_FIELD_NOT_RESTORED", justification = "Specific method to restore it")
     private transient SuiteResult parent;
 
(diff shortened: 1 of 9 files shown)


jjhiblot:
Thanks for your prompt review and merge.

jjhiblot:
I am not sure that it is part of the "classic" JUnit. it used by some test automation tools however.

mem89de:
This Pull Request blew up the build for our legacy monolith with more than 50k unit tests. Is there a way to disable the properties?

jjhiblot:
That is unfortunate. No there is no way to disable it. 
Maybe we could add a kind of filter in the configuration of the plugin.
@timja I'm willing to do it but I would need a few pointers. I'm think of having a regex to include and one to exclude the properties based on their name.

mem89de:
This would be very nice :)

Is there a quick workaround for this problem, that comes to your mind @jjhiblot? Is there maybe a way to disable the creation of the properties nodes during testing itself?

Otherwise there is only the downgrade of the plugin.

timja:
> This Pull Request blew up the build for our legacy monolith with more than 50k unit tests. Is there a way to disable the properties?

how does it blow it up? what issue is it causing?

timja:
@jjhiblot can you add a field that people have to use to enable properties?

If you can't do it soon we can revert the PR and then re-enable at a later date.

jjhiblot:
@timja I'll try to work on getting a boolean in the parameters tomorrow. Something like a "keepProperties" boolean option, in the same way as 'keepLongStdio"

timja:
👉 https://github.com/jenkinsci/junit-plugin/pull/546

PR #546, opened 2023-07-04
Add a "keepProperties" option

jjhiblot:
If 'false' the properties attached to a test suite or a test case are ignored.
By default the properties are ignored.


timja (approved):
thanks!

mem89de:
Thank you so much! We will shedule to install the update as soon as it is released :)

timja:
Released now

mem89de:
We have installed the plugin on friday. It seems, everything is back to normal. Thank you!

PR #555, opened 2023-08-04
[JENKINS-59846] Hyperlink URLs in JUnit Test Report

julieheard:
Some history on this ticket - for [SECURITY-2888](https://issues.jenkins.io/browse/SECURITY-2888) a while back, I removed the ability to click on the links in the test report. The original annotate method allowed unsafe links to be clickable, it was unsafe because it allowed anything to be passed in and annotated as a link.

In this PR what I have done is revert the removal of clickable links, I have added some security on top in the form of using the [OWASP Java HTML Sanitizer](https://owasp.org/www-project-java-html-sanitizer/). This enables the links to be clickable but sanitises them, so nothing nasty can get passed through.

### Testing done
I have run mvn verify locally and all working. Also I have manually tested this it againts the original vulnerability to make sure it still blocks unsafe links tso you can't trigger anthing unsafe.

I would really appreciate some more testing particularly from @Ingo987 @academic-sakharov 🙂

Before:
image

After:
image

Testing running with a malicious report with `https://google.com\:quot;onclick=alert(1)` as the test URL:
image
You can see the tiny escaped URL at the bottom 

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,11 @@
             <artifactId>pipeline-utility-steps</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+            <artifactId>owasp-java-html-sanitizer</artifactId>
+            <version>20220608.1</version>
+        </dependency>
     </dependencies>
     <dependencyManagement>
         <dependencies>
--- a/src/main/java/hudson/tasks/test/TestResult.java
+++ b/src/main/java/hudson/tasks/test/TestResult.java
@@ -23,13 +23,15 @@
  */
 package hudson.tasks.test;
 
-import hudson.tasks.junit.TestAction;
 import hudson.model.Run;
 import hudson.model.Result;
 
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+
+import org.owasp.html.HtmlPolicyBuilder;
+import org.owasp.html.PolicyFactory;
 
 import static java.util.Collections.emptyList;
 
@@ -276,7 +278,8 @@ public abstract class TestResult extends TestObject {
     }
 
     /**
-     * Annotate some text -- what does this do? 
+     * Annotate text by turning it into a clickable link and then sanitizing it using owasp-java-html-sanitizer
+     * @see <a href="https://www.javadoc.io/doc/com.googlecode.owasp-java-html-sanitizer/owasp-java-html-sanitizer/20160628.1/org/owasp/html/HtmlPolicyBuilder.html">owasp HtmlPolicyBuilder javadoc</a>
      * @param text Text to use to annotate the actions.
      *
      * @return the provided text HTML-escaped.
@@ -284,11 +287,16 @@ public abstract class TestResult extends TestObject {
     public String annotate(String text) {
         if (text == null)
                 return null;
-        text = text.replace("&", "&amp;").replace("<", "&lt;");
 
-        for (TestAction action: getTestActions()) {
-                text = action.annotate(text);
-        }
-        return text;
+        text = text.replace("&", "&amp;").replace("<", "&lt;").replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
+
+        PolicyFactory policy = new HtmlPolicyBuilder()
+                .allowElements("a")
+                .allowUrlProtocols("https")
+                .allowUrlProtocols("http")
+                .allowAttributes("href").onElements("a")
+                .toFactory();
+
+        return policy.sanitize(text);
     }
 }
--- a/src/main/java/hudson/tasks/test/TestResult.java
+++ b/src/main/java/hudson/tasks/test/TestResult.java
@@ -45,6 +45,13 @@ import static java.util.Collections.emptyList;
  */
 public abstract class TestResult extends TestObject {
 
+    private static final PolicyFactory POLICY_DEFINITION= new HtmlPolicyBuilder()
+            .allowElements("a")
+            .allowUrlProtocols("https")
+            .allowUrlProtocols("http")
+            .allowAttributes("href").onElements("a")
+            .toFactory();;
+
     /**
      * If the concept of a parent action is important to a subclass, then it should
      * provide a non-noop implementation of this method. 
@@ -290,13 +297,6 @@ public abstract class TestResult extends TestObject {
 
         text = text.replace("&", "&amp;").replace("<", "&lt;").replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
 
-        PolicyFactory policy = new HtmlPolicyBuilder()
-                .allowElements("a")
-                .allowUrlProtocols("https")
-                .allowUrlProtocols("http")
-                .allowAttributes("href").onElements("a")
-                .toFactory();
-
-        return policy.sanitize(text);
+        return POLICY_DEFINITION.sanitize(text);
     }
 }
--- a/src/main/java/hudson/tasks/test/TestResult.java
+++ b/src/main/java/hudson/tasks/test/TestResult.java
@@ -296,7 +296,7 @@ public abstract class TestResult extends TestObject {
         if (text == null)
                 return null;
 
-        text = text.replace("&", "&amp;").replace("<", "&lt;").replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
+        text = text.replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
 
         for (TestAction action: getTestActions()) {
             text = action.annotate(text);
--- a/src/main/java/hudson/tasks/test/TestResult.java
+++ b/src/main/java/hudson/tasks/test/TestResult.java
@@ -23,6 +23,7 @@
  */
 package hudson.tasks.test;
 
+import hudson.tasks.junit.TestAction;
 import hudson.model.Run;
 import hudson.model.Result;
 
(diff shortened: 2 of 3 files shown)


src/main/java/hudson/tasks/test/TestResult.java
```diff
@@ -276,19 +278,25 @@ public String toPrettyString() {
...
                 return null;
-        text = text.replace("&", "&amp;").replace("<", "&lt;");
 
-        for (TestAction action: getTestActions()) {
-                text = action.annotate(text);
-        }
-        return text;
+        text = text.replace("&", "&amp;").replace("<", "&lt;").replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
+
+        PolicyFactory policy = new HtmlPolicyBuilder()
```
jtnord, line 293: `POlicyFactory` is marked as `ThreadSafe` do we really want to create a new one every time this is called?

src/main/java/hudson/tasks/test/TestResult.java
```diff
@@ -276,19 +278,25 @@ public String toPrettyString() {
...
     public String annotate(String text) {
         if (text == null)
                 return null;
-        text = text.replace("&", "&amp;").replace("<", "&lt;");
 
-        for (TestAction action: getTestActions()) {
-                text = action.annotate(text);
-        }
-        return text;
+        text = text.replace("&", "&amp;").replace("<", "&lt;").replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
```
Kevin-CB, line 291: Could you add a test to confirm that `https://javascript:alert(1)` getting transformed into `<a href="javascript:alert(1)"` will correctly be caught by the policy?

src/test/java/hudson/tasks/junit/CaseResultTest.java
```diff
@@ -98,24 +98,23 @@ public void testIssue20090516() throws Exception {
...
-                     "https://nowhere.net/");
+                     "<a href=\"https://nowhere.net/\">https://nowhere.net/</a>");
         assertOutput(cr,"stuffhttp://nowhere.net/", "stuffhttp://nowhere.net/");
         assertOutput(cr,"a < b && c < d", "a &lt; b &amp;&amp; c &lt; d");
         assertOutput(cr,"see <http://nowhere.net/>",
-                     "see &lt;http://nowhere.net/>");
+                     "see &lt;<a href=\"http://nowhere.net/\">http://nowhere.net/</a>&gt;");
         assertOutput(cr,"http://google.com/?q=stuff&lang=en",
-                     "http://google.com/?q=stuff&amp;lang=en");
+                     "<a href=\"http://google.com/?q&#61;stuff&amp;lang&#61;en\">http://google.com/?q&#61;stuff&amp;lang&#61;en</a>");
```
jtnord, line 115: this looks questionable - should it not be `<a href=\"http://google.com/?q=stuff&lang=en\">http://google.com/?q&#61;stuff&amp;lang&#61;en</a>

src/main/java/hudson/tasks/test/TestResult.java
Kevin-CB, line 291: Something like `https://google.com"onclick=alert(1)"` won't be transformed into `<a href=\"https://google.com" onclick=alert(1) ""`?

But I guess it won't work due to the policy

pom.xml
```diff
@@ -194,6 +194,11 @@
             <artifactId>pipeline-utility-steps</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
```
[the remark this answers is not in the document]
daniel-beck: Seems slightly awkward to have a plugin dependency to only get to the library, when it's not an actual library plugin. Unsure.

(I would discourage using markup formatter plugin functionality, otherwise behavior of this feature depends on what's allowed for user-provided markup. That caused some trouble in uno-choice in the past.)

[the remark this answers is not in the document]
jtnord: > Seems slightly awkward to have a plugin dependency to only get to the library, when it's not an actual library plugin. Unsure.

it is at least what [timestamper](https://github.com/jenkinsci/timestamper-plugin/blob/ce7a5ac5422af70ab9d96638e333f698137c6920/src/main/java/hudson/plugins/timestamper/format/TimestampFormatUtils.java#L17-L22) [does](https://github.com/jenkinsci/timestamper-plugin/blob/ce7a5ac5422af70ab9d96638e333f698137c6920/pom.xml#L78-L81) :-/

so there are now 3 plugins that use the library which is normally the time to create an API plugin then...?
The security aspects of that esp for the markup formatter make me a little squeamish though...

src/main/java/hudson/tasks/test/TestResult.java
julieheard: Thank you for your comments 😄 
I have added two more tests to test the unsafe characters to make it obvious that these chars are escaped

src/main/java/hudson/tasks/test/TestResult.java
julieheard: Done, I have moved the policy factory outside of the method so there is only 1 created for the class

src/test/java/hudson/tasks/junit/CaseResultTest.java
julieheard: Equals signs are getting escaped, we could easily put them back in after it has been sanitised with a replaceAll but then we are kind of undoing the sanitising 😕

pom.xml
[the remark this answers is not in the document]
julieheard: So conclusion here is to not rely on the markup-formatter, but instead to create another jenkins ticket to create a API plugin? When that is up and sorted we can switch other plugins over to using the API plugin. Give me a thumbs up on this and I will create the jira ticket 🙂

src/main/java/hudson/tasks/test/TestResult.java
```diff
@@ -276,19 +286,22 @@ public String toPrettyString() {
...
      * @param text Text to use to annotate the actions.
      *
      * @return the provided text HTML-escaped.
      */
     public String annotate(String text) {
         if (text == null)
                 return null;
-        text = text.replace("&", "&amp;").replace("<", "&lt;");
+
+        text = text.replaceAll("\\b(https?://[^\\s)>]+)", "<a href=\"$1\">$1</a>");
```
jtnord, line 299: text is essentially user supplied.

have you thought about if it contained non allowed HTML ( e.g. `this is <strong>some text</strong> in an assertion that failed`   
what happens in this case - does the message loose fidelity by being interpretted as pseudo HTML?
(is it displayed as the litteral "this is <strong>some text</strong> in an assertion that failed" or the litteral "this is some text in an assertion that failed" or something else?

I would semi expect the sanitizer to throw away the `strong` part - but what it really wanted is to see the underlying raw text here. (that is I expect to see in the web UI the `<strong>` part.   but I have a feeling this will be thrown away.

src/test/java/hudson/tasks/junit/CaseResultTest.java
```diff
@@ -119,6 +119,9 @@ public void testIssue20090516() throws Exception {
                      "<a href=\"https://google.com\">https://google.com&#34;onclick&#61;alert(1</a>)&#34;");
         assertOutput(cr,"unsafe characters are = \" ' < > &",
                      "unsafe characters are &#61; &#34; &#39; &lt; &gt; &amp;");
+        assertOutput(cr,"this is <strong>some text</strong> in an assertion that failed",
+                     "this is some text in an assertion that failed");
```
jtnord, line 123: I would consider this a regression - if the output is munged in any way that what I see on screen is not what was the output then I would be at a loss as to what the error could be.

Consider a failure where 

```java 
String expected = "some text"
assertThat("some <strong>text</strong>, is(expected));
```

the error would be like `"failure expected "some text" but got "some text"`  and that is just plain wrong.

jtnord: the text could even be non HTML...

`7 < 8`  would the sanitizer braf here - and strip the 8 - would it cause rendering issues when displayed in Jenkins.

julieheard: Yes I see your point here, that would be confusing. I have removed that last test. We already have one which tests `a < b && c < d` which outputs `a &lt; b &amp;&amp; c &lt; d`

pom.xml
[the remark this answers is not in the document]
jtnord: I don't think we have a consensus :)

src/main/java/hudson/tasks/test/TestResult.java
jtnord: FWIW that screenshot shows you get different output in different places - in on the error message has "strong" and in the test overview page it does not...    😱

julieheard:
Converted to a draft as I am going to play around with different sanitising methods. 

I have just tried using org.jsoup.safety.Safelist; and it has the same issue that it removes the tags completely.

src/test/java/hudson/tasks/junit/CaseResultTest.java
julieheard: I have spent some time playing with this url sanitizing and there is not a method that I can find that sanitizes the output without removing forbidden tags. This is unless we want to write a whole new sanitizing plugin ourselves. Looking at the plugin, one of the jelly pages annotates the error detail and the other renders it in plaintext.

In terms of a compromise, we could output the error details in two forms, the HTML sanitized as well as the plaintext, like this:
image
image

This gets you the link but might look a bit messy. We could also add more comments to the method to say more explicitly that it removes the unsafe tags. 

I think at the moment this is the only sensible-ish answer, which still leaves this as a regression as the annotate method will adapt the data fed into it to remove parts.

Ingo987:
Hi Julie,

What is the conclusion for this enhancement? Just tried latest released Junit version and the problem is still there. Probably I misinterpreted “closed” – I was hoping it is “fixed”.
Regards,
Ingo

Ingo987:
Hi Julie,

thanks for coming back on this.
In fact, I built my own version of the JUnit plugin now and was using that to keep up w/ the other changes to Jenkins (esp. the transition to RH8)

So I'm happy for now, but for the long run it would be great to see this change getting productive.

Regards,
Ingo
________________________________

Issue #570, opened 2023-09-18
Stacktrace after moving Junit results to external storage and cannot load graph of previous results


jonesbusy:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.423
OS: Linux - 6.2.0-32-generic
Java: 11.0.20 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)
---
analysis-model-api:11.7.1
antisamy-markup-formatter:162.v0e6ec0fcfcf6
apache-httpcomponents-client-4-api:4.5.14-208.v438351942757
bootstrap5-api:5.3.0-1
bouncycastle-api:2.29
branch-api:2.1128.v717130d4f816
caffeine-api:3.1.8-133.v17b_1ff2e0599
checks-api:2.0.1
cloudbees-folder:6.848.ve3b_fd7839a_81
commons-lang3-api:3.13.0-62.v7d18e55f51e2
commons-text-api:1.10.0-78.v3e7b_ea_d5a_fe1
config-file-provider:953.v0432a_802e4d2
configuration-as-code:1700.v6f448841296e
credentials:1271.v54b_1c2c6388a_
credentials-binding:636.v55f1275c7b_27
customizable-header:18.v820da_cd15cb_1
dark-theme:372.v79b_02c754b_29
data-tables-api:1.13.5-1
database:191.vd5981b_97a_5fa_
database-postgresql:81.v136de0a_08ef9
display-url-api:2.3.9
durable-task:523.va_a_22cf15d5e0
echarts-api:5.4.0-5
font-awesome-api:6.4.0-2
forensics-api:2.3.0
git:5.2.0
git-client:4.5.0
git-parameter:0.9.19
github:1.37.3
github-api:1.314-431.v78d72a_3fe4c3
github-branch-source:1740.v51d5810e9e8c
inline-pipeline:1.0.3
instance-identity:173.va_37c494ec4e5
ionicons-api:56.v1b_1c8c49374e
jackson2-api:2.15.2-350.v0c2f3f8fc595
jakarta-activation-api:2.0.1-3
jakarta-mail-api:2.0.1-3
javax-activation-api:1.2.0-6
javax-mail-api:1.6.2-9
jaxb:2.3.8-1
jjwt-api:0.11.5-77.v646c772fddb_0
jquery3-api:3.7.0-1
junit:1217.v4297208a_a_b_ce
junit-sql-storage:225.v96c624d82841
[...]

PR #571, opened 2023-09-18
Fix chart trend after moving from default storage to external storage

jonesbusy:
This fixes chart trend after moving from default storage to external storage

Fixes https://github.com/jenkinsci/junit-plugin/issues/570

Instead of loading forever just return an empty trends chart.

### Testing done

Create build with local storage
Switch to external storage
Check that there is no exception

**Before**

junit_plugable_storage

**After**

junit_storage_1

Triggering more build continue to rebuild the trend chart

junit_storage_2

```[tasklist]
### Submitter checklist
```

The situation is not ideal but I think is better than before.

Regards,


timja (approved):

timja:
Thanks!

PR #596, opened 2024-01-03
Add start time fields and logic for setting start time of results in epoch milliseconds

bheleker:
…s, test packages, and test suites

This pull request is to add information about start times to the different test result classes. Start times are recorded using epoch milliseconds. Many tests in JUnit (and even more converted to JUnit via other plugins i.e. xUnit) include start times for test suites as well as test cases which could be useful information for third-party plugins to have. The default value is -1 for an unset start time and the package start time is the earliest start time of its test cases. Additionally, the overall test result start time is the earliest set start time of a SuiteResult. The largest adds are the parseTime method for parsing the epoch millisecond time from a timestamp and the logic for setting start times. It is intended to be as non-invasive as possible, so no logic is introduced to derive the start time of a test class, but that can be derived by those that want to use the start times. This introduces no breaking changes and no change to the frontend.  It will not affect the performance or output of the plugin itself and only introduces more possible information for those extending the plugin. 

### Testing done

No unit tests were added for the change. However, I made the change alongside my development of an extended TestDataPublisher for pseudo-tracing and metric data collection and was able to verify that start time for suites and test cases matched what was parsed from their reports. Additionally, all unit tests are still passing as the parsing of a start time does not affect the plugin's performance. Omission of a start time simply means it is set to -1 which I verified when testing my extension. Additionally, inclusion of a non-ISO start time results in a LOGGER.warning but does not affect the program's ability to run.

```[tasklist]
### Submitter checklist
- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x ] Ensure that the pull request title represents the desired changelog entry
- [x ] Please describe what you did
- [x ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -265,7 +265,7 @@ public final class SuiteResult implements Serializable {
         }
 
         // offset for start time of cases if none is case timestamp is not specified
-        double caseStartOffset = 0;
+        long caseStartOffset = 0;
         List<Element> testCases = suite.elements("testcase");
         for (Element e : testCases) {
             // https://issues.jenkins-ci.org/browse/JENKINS-1233 indicates that
@@ -297,8 +297,8 @@ public final class SuiteResult implements Serializable {
             }
             // Else estimate start time using sum of previous case durations in suite
             else if (startTime != -1) {
-                caze.setStartTime((long)(startTime + caseStartOffset));
-                caseStartOffset += caze.getDuration();
+                caze.setStartTime(startTime + caseStartOffset);
+                caseStartOffset += (long)(caze.getDuration() * 1000);
             }
         }
 
--- a/src/main/java/hudson/tasks/junit/SuiteResult.java
+++ b/src/main/java/hudson/tasks/junit/SuiteResult.java
@@ -41,6 +41,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.Serializable;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -73,6 +74,7 @@ public final class SuiteResult implements Serializable {
     private final String stdout;
     private final String stderr;
     private float duration;
+    private long startTime;
     private final Map<String, String> properties;
 
     /**
@@ -242,6 +244,12 @@ public final class SuiteResult implements Serializable {
             this.enclosingBlocks.addAll(pipelineTestDetails.getEnclosingBlocks());
             this.enclosingBlockNames.addAll(pipelineTestDetails.getEnclosingBlockNames());
         }
+        
+        // check for timestamp attribute and set start time if present
+        if(timestamp != null && !timestamp.equals("")) 
+            this.startTime = parseTime(timestamp);
+        else
+        	this.startTime = -1;
 
         // check for test suite time attribute
         if ((this.time = suite.attributeValue("time")) != null) {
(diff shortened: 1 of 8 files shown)


src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -242,6 +244,12 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, boolea
             this.enclosingBlocks.addAll(pipelineTestDetails.getEnclosingBlocks());
             this.enclosingBlockNames.addAll(pipelineTestDetails.getEnclosingBlockNames());
         }
+        
+        // check for timestamp attribute and set start time if present
+        if(timestamp != null && !timestamp.equals(""))
```
timja, line 249: add braces around if else

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -242,6 +244,12 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, boolea
             this.enclosingBlocks.addAll(pipelineTestDetails.getEnclosingBlocks());
             this.enclosingBlockNames.addAll(pipelineTestDetails.getEnclosingBlockNames());
         }
+        
+        // check for timestamp attribute and set start time if present
+        if(timestamp != null && !timestamp.equals("")) 
+            this.startTime = parseTime(timestamp);
+        else
+        	this.startTime = -1;
```
timja, line 252: switch tab to spaces

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -274,8 +284,19 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, boolea
...
-
-            addCase(new CaseResult(this, e, classname, keepLongStdio, keepProperties));
+            
+            CaseResult caze;
+            addCase(caze = new CaseResult(this, e, classname, keepLongStdio, keepProperties));
+            
+            // If timestamp is present for <testcase> set startTime of new CaseResult.
+            String caseStart = e.attributeValue("timestamp");
+            if(caseStart != null && !caseStart.equals(""))
+            		caze.setStartTime(parseTime(caseStart));
```
timja, line 294: tabs to spaces

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -274,8 +284,19 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, boolea
...
+            // Else estimate start time using sum of previous case durations in suite
+            else if(startTime != -1) {
```
timja, line 296:
```suggestion
            else if (startTime != -1) {
```

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -274,8 +284,19 @@ private SuiteResult(File xmlReport, Element suite, boolean keepLongStdio, boolea
...
+            	caze.setStartTime((long)(startTime + caseStartOffset));
+            	caseStartOffset += caze.getDuration();
```
timja, line 298: tabs to spaces

src/main/java/hudson/tasks/junit/SuiteResult.java
```diff
@@ -495,6 +524,25 @@ public void setParent(hudson.tasks.junit.TestResult parent) {
             c.freeze(this);
         return true;
     }
+    
+    /**
+     * Parses time as epoch milli from time string
+     * @param time
+     * @return time in epoch milli
+     */
+    public long parseTime(String time) {
```
timja, line 533: please add a unit test for this

timja (commented):
generally looks fine, some formatting issues need fixing and can you please add some test coverage

bheleker:
> generally looks fine, some formatting issues need fixing and can you please add some test coverage

I've fixed the formatting and added unit tests for parseTime. When I have time I will add some tests for startTime after parsing xml.

Thanks

bheleker:
I have completed updates to startTime logic and added unit tests for the parseTime function and the start times from a jUnit test result file

timja (approved):

timja:
not manually tested

Issue #219, opened 2020-12-02
...[truncated 135494 chars]...


brianjmurrell:
I'm not making much sense of the explanation of this option:

**Retain long standard output/error**: If checked, any standard output or error from a test suite will be retained in the test results after the build completes. (This refers only to additional messages printed to console, not to a failure stack trace). _Such output is always kept if the test failed_, but by default lengthy output from passing tests is truncated to save space...

Is that saying that that option **only** applies to passing tests and that failing tests should **always** have full output and not have `...[truncated 135494 chars]...` in the middle of it's **Standard Output** section in the JUnit results?

IOW, I should **never** see `...[truncated 135494 chars]...` in a failing test result, correct?  If correct, then this issue is to report that I do see that even in failing test results.

brianjmurrell:
I should add, and maybe this is a separate ticket, but it's super annoying that the `tail` after the `...[truncated 135494 chars]...` is displayed literally instead of interpreting the `\n`s:

```
21:40:01 DEBUG| [stdout] [1,0]<stdout>:Instance 1: starting format of SCM (dcpm:/mnt/daos1)
21:40:01 DEBUG| [stdout] [1,0]<stdout>:DEBUG 21:40:01.556730 instance_storage.go:107: /mnt/daos0 (dcpm) needs format: true
21:40:01 DEBUG| [stdout] [1,0]<stdout>:Instance 0: starting format of SCM (dcpm:/mnt/daos0)
21:40:12 DEBUG| [stdout] [1,3]<stdout>:Instance 0: finished format of SCM (dcpm:/mnt/daos0)
21:40:12 DEBUG| [stdout] [1,2]<stdout>:Instance 0: finished format of
...[truncated 135494 chars]...
[...]

PR #601, opened 2024-01-23
[JENKINS-27931] Replace keepLongStdio with stdioRetention, allow test output to be retained for failed tests only

C0urante:
This PR deprecates the `keepLongStdio` property and replaces it with `stdioRetention`. The behavior for these properties (✅ indicates long output retention; a blank box represents long output truncation) is:

| | All tests passed | At least one test failed |
| - | - | - |
| (default) | | |
| `keepLongStdio: false` | | |
| `keepLongStdio: true` | ✅ | ✅ |
| `stdioRetention: 'none'` | | |
| `stdioRetention: 'failed'` | | ✅ |
| `stdioRetention: 'all'` | ✅ | ✅ |

Note that this does not apply when test output is redirected to a separate file (i.e., [this code path](https://github.com/jenkinsci/junit-plugin/blob/14bf914456c49025bb28c931dc0304521af2edc1/src/main/java/hudson/tasks/junit/SuiteResult.java#L315) is triggered). For that scenario, existing behavior is mostly retained; the only difference is that the 1MB cap on test output (which is currently allowed if `keepLongStdio = true`) is also allowed if there are test failures and `stdioRetention` is set to `failed`, or if `stdioRetention` is set to `all`.

Relevant links:
- [JENKINS-27931](https://issues.jenkins.io/browse/JENKINS-27931)
- GitHub issue 219: #219 

#### Automated testing

Unit testing coverage is expanded to cover these cases:
- Input is truncated when `stdioRetention = failed` and the test(s) have passed
- Input is truncated when `stdioRetention = none` and test(s) have failed
- Input is not truncated when 'stdioRetention = failed' and test(s) have failed
- Input is not truncated when `stdioRetention = all` and test(s) have passed

#### Manual testing

- Manual testing covered all possible values of `keepLongStdio` and `stdioRetention`, as well as when no value is provided for either

```[tasklist]
### Submitter checklist
- [X] Link to relevant issues in GitHub or Jira
- [X] **(N/A)** Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```


src/main/java/hudson/tasks/junit/StdioRetention.java
```diff
@@ -0,0 +1,15 @@
+package hudson.tasks.junit;
+
+public enum StdioRetention {
+
+    all,
+    failed,
+    none;
```
timja, line 7: normal code convention in Java is for enums to be in all uppercase letters.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
```diff
@@ -30,8 +30,12 @@ THE SOFTWARE.
...
         <f:textbox />
     </f:entry>
-    <f:entry field="keepLongStdio" title="">
-        <f:checkbox name="keepLongStdio" title="${%Retain long standard output/error}"/>
+    <f:entry field="stdioRetention" title="">
+        <select name="stdioRetention">
+            <option value="none">None</option>
+            <option value="failed">Failed tests only</option>
+            <option value="all">All</option>
+        </select>
```
timja, line 38: these should be fetched from code with a `doFillXXX` method using an `f:select`.

timja (commented):
not tested locally, can you check the failing tests please?

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -278,12 +288,11 @@ static String possiblyTrimStdio(Collection<CaseResult> results, boolean keepLong
...
-            if (result.errorStackTrace != null) {
-                return HALF_MAX_FAILING_SIZE;
-            }
-        }
-        return HALF_MAX_SIZE;
+        return hasFailures(results) ? HALF_MAX_FAILING_SIZE : HALF_MAX_SIZE;
+    }
+
+    private static boolean hasFailures(Collection<CaseResult> results) {
+        return results.stream().anyMatch(r -> r.errorStackTrace != null);
```
jglick, line 295: IIUC this is just switching to `Stream`, but not changing method behavior.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -244,14 +251,17 @@ static String fixNULs(String stdio) { // JENKINS-71139
...
+     * Flavor of {@link #possiblyTrimStdio(Collection, StdioRetention, String)} that doesn't try to read the whole thing into memory.
      */
     @SuppressFBWarnings(value = "DM_DEFAULT_ENCODING", justification = "Expected behavior")
-    static String possiblyTrimStdio(Collection<CaseResult> results, boolean keepLongStdio, File stdio) throws IOException {
+    static String possiblyTrimStdio(Collection<CaseResult> results, StdioRetention stdioRetention, File stdio) throws IOException {
         long len = stdio.length();
-        if (keepLongStdio && len < 1024 * 1024) {
+        boolean keepAll = stdioRetention == StdioRetention.all
+                || (stdioRetention == StdioRetention.failed && hasFailures(results));
+        if (keepAll && len < 1024 * 1024) {
```
jglick, line 261: So this is not addressing #219: there is still a hard cap on stdio size.

jglick (commented):
It is hard for me to follow exactly what this is changing. It would be helpful to have a table summarizing the matrix of possibilities:

* passing vs. failing
* old `!keepLongStdio` vs. `keepLongStdio` vs. new `all` vs. `failed` vs. `none`
* stdio < 1Kb vs. 1Kb < stdio < 1Mb vs. stdio > 1Mb

And perhaps there needs to be a way to adjust the size thresholds on a per-step basis. I have certainly come across cases where a test failed yet noisy output went beyond 1Mb and the critical diagnostic information was in the middle; yet on the other hand if the output were 10Mb I would be afraid of the impact of storing it.

src/main/java/hudson/tasks/junit/StdioRetention.java
C0urante: The intent was to permit lowercase values. Any idea if there's a way to enable case-insensitive parsing for enum types with `@DataBoundSetter` methods?

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
C0urante: I've taken a stab at this; let me know how it looks. I'm also not terribly clear on how to test this (I've been using a `Jenkinsfile` to configure the plugin for testing); if there's a resource out there on how to validate this, please let me know.

src/main/java/hudson/tasks/junit/CaseResult.java
C0urante: Yeah, that's correct. If you'd prefer, we can stick to the existing for-loop style.

src/main/java/hudson/tasks/junit/CaseResult.java
C0urante: That's true, the intention was to preserve existing behavior by making `stdioRetention = 'all'` completely identical to `keepLongStdio = true`.

I'd love to include this scenario as well and allow long output (even beyond 1MB) to be preserved when it's hit, but that may bite users who upgrade and all of a sudden find that massive files are being stored after their CI runs.

So to be clear, this PR does still address #219, just not in every possible case.

src/main/java/hudson/tasks/junit/CaseResult.java
jglick: No opinion (nor am I a maintainer), just noting while trying to follow what the code changes do.

src/main/java/hudson/tasks/junit/CaseResult.java
jglick: > does still address #219, just not in every possible case

Does it?

> I should never see `...[truncated 135494 chars]...` in a failing test result […yet] I do see that even in failing test results.

which AFAICT is still true after this PR.

Still just trying to understand what this is PR is and is not changing. The new table does not show the effect on different stdio sizes, or maybe I am just not getting it. (@timja any clearer to you?)

C0urante: > Does it?

Yes, and I have personally verified this with manual testing.

> which AFAICT is still true after this PR.

If we're taking this literally and focusing on the word "never", then no, it doesn't. But like I noted before, this PR does prevent truncation from happening in some scenarios. See [here](https://github.com/C0urante/junit-plugin/blob/c34254c05e814dbe87dbe733ef41879ffe8609b8/src/main/java/hudson/tasks/junit/CaseResult.java#L231-L247).

File sizes only come into play with a [different variant](https://github.com/C0urante/junit-plugin/blob/c34254c05e814dbe87dbe733ef41879ffe8609b8/src/main/java/hudson/tasks/junit/CaseResult.java#L231-L247) of the `CaseResult::possiblyTrimStdio` method, which I've intentionally refrained from making more invasive changes to.

C0urante:
Thank you both for the rapid turnaround!

@jglick I've tweaked the description of the PR to help shed some more light on the change in behavior (I know it doesn't match the exact format you requested, but I couldn't think of a simple way to present a 3D matrix in GitHub markdown--please let me know if anything is still unclear).

I do agree that it'd be nice to be able to tweak the truncation size, but I'm hoping that can come as a follow-up item. This change in behavior is completely opt-in and shouldn't make it any easier for people to shoot themselves in the foot than it already is, and I know that at least for my use case, 2.5MB log files for failing tests are not uncommon and it's been a bit of a struggle to diagnose flaky tests without access to them.

src/main/java/hudson/tasks/junit/CaseResult.java
jglick: So if I am using https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html which writes `*-output.txt`, and I have a test which writes out 2Mb and output and then fails, I am still going to see the middle half truncated. Right?

C0urante:
@timja I've fixed the failing tests and this should be ready for another round.

src/main/java/hudson/tasks/junit/CaseResult.java
C0urante: Only if the system property `maven.test.redirectTestOutputToFile`, which defaults to false, is set to true.

jglick: As in for example https://github.com/jenkinsci/plugin-pom/blob/e4a9d6957748be5cc3e1c3026014c3db60047d7b/pom.xml#L1030-L1040.

I guess the right fix is to stop copying stdio into `junitResult.xml` in the built-in storage, instead writing/copying it to a separate file in the build directory, and then reading from that on demand in display threads. This would allow the truncation hack to be deleted without risking memory problems in the controller. What I am unsure of is whether there is then a use case for ever _not_ retaining stdio, even for passing tests. Disk space?

jglick: > this approach potentially painting us into a corner

I do not think so, just trying to think through the implications.

C0urante: I'm going to mark this resolved as I believe I've addressed your concerns.

src/main/java/hudson/tasks/junit/StdioRetention.java
C0urante: Pushed a new commit that does this work manually; happy to change things if there's a cleaner approach.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/config.jelly
C0urante: Never mind, figured out how to test (configure -> Post-build Actions -> Publish JUnit test result report).

src/main/java/hudson/tasks/junit/StdioRetention.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Locale;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum StdioRetention {
+
+    ALL("All"),
+    FAILED("Failed tests only"),
+    NONE("None");
```
timja, line 11: These shouldn't be hardcoded in english, they need to use a messages class, see https://www.jenkins.io/doc/developer/internationalization/i18n-source-code/

C0urante: Thanks, done. I've also added help info for the `stdioRetention` option and added a deprecation notice for `keepLongStdio`. I've only added/altered English variants though. Do we need to add/alter German, Japanese, and Chinese info in this PR as well, or can that be addressed as a follow-up?

src/main/java/hudson/tasks/junit/StdioRetention.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public static final StdioRetention DEFAULT = NONE;
+
+    public static StdioRetention fromKeepLongStdio(boolean value) {
+        return value ? ALL : NONE;
+    }
+
+    public static StdioRetention parse(String value) {
+        if (value == null)
+            return DEFAULT;
```
timja, line 31: braces around if statements
```suggestion
        if (value == null) {
            return DEFAULT;
        }
```

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-keepLongStdio.html
```diff
@@ -1,4 +1,5 @@
 <div>
+   (Deprecated; use stdioRetention instead)
```
timja, line 2: probably something like this to emphasise it:
```suggestion
  <strong>Deprecated, use stdioRetention instead.</strong>
  <p>
```

src/main/java/hudson/tasks/junit/StdioRetention.java
timja: no need to add them if you aren't a native speaker.

the normal convention is if you change an english localised string you clear out the translations.
if you remove an english localised string you remove all translations

timja (approved):
untested but seems sensible, thanks for the work on this.

C0urante:
Thanks @timja

alecharp:
This is breaking the test https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/blob/9cb23efd4a61e4299763d599e8552207cf65ada9/src/test/java/org/jenkinsci/plugins/junitrealtimetestreporter/RealtimeJUnitStepTest.java#L210 in `junit-realtime-test-reporter`. That test seems to be adapted to this fix.

timja:
@C0urante would you be able to take care of that?

C0urante:
@timja @alecharp Yep, on it: https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/pull/149

Going to look into the declarative pipeline migration assistant plugin next.

src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
```diff
@@ -507,5 +520,26 @@ public FormValidation doCheckHealthScaleFactor(@QueryParameter double value) {
                     (int) (100.0 - Math.max(0.0, Math.min(100.0, 5 * value)))
             ));
         }
+
+        public ListBoxModel doFillStdioRetentionItems(@QueryParameter("stdioRetention") String value) {
```
jglick, line 524: This is not exposed from the **Pipeline Syntax** page, which just shows an empty dropdown (with no help); I think you forgot to delegate to this from `JUnitResultsStep.DescriptorImpl`.

src/main/resources/hudson/tasks/junit/JUnitResultArchiver/help-stdioRetention.html
```diff
@@ -0,0 +1,18 @@
...
+    always retained for tests, but by default, excessively long output is truncated
+    to save disk space. This property can be used to control whether that truncation
+    takes place:
+    <ul>
+        <li>If set to 'all', then no truncation will take place and the complete output from
+        every test will be stored.</li>
+        <li>If set to 'failed', then the complete output from all failed tests will be stored,
+        but long output from passing tests will be truncated.</li>
+        <li>If set to 'none', then long output from every test will be truncated, regardless
+        of whether they passed or failed.</li>
```
jglick, line 14: Uppercase seems to be the canonical form advertised by `doFillStdioRetentionItems` even if lowercase is also tolerated?

PR #610, opened 2024-03-26
Allow PCT testing with `database-sqlite`

basil:
Discovered in https://github.com/jenkinsci/bom/pull/3057 when trying to add `database-sqlite` to BOM. Both SQLite and H2 database plugins attempt to set themselves as the default, and when SQLite wins then this test fails. Fixed by changing the test to not make any assumptions about the default and instead explicitly set the database to the one the test expects.

### Testing done

Tested in context in PCT. Passes after this PR.


timja (approved):
if it works 🤷

PR #622, opened 2024-06-15
Add flow node ID to external storage API

timja:
This is currently a breaking change although it could likely be made more compatible.
There's only realtime junit to adapt anyway.

I'm not sure this is the best approach and would appreciate any guidance.

See https://github.com/jenkinsci/junit-sql-storage-plugin/pull/414/files#diff-1202c93b95ce3f5252dc25bf1a5ab73d75ebc67052d675c97d5e146a3448596aR140

In order to link the traces that are being used for instrumenting the junit sql plugin I need access to the `FlowNodeMonitoringAction` to get the open telemetry trace parent id.

That is attached to the flow node and I couldn't find a way other than passing it down to get to it.
The normal `Context.current` from otel doesn't work as the otel plugin uses actions for tracking and isn't able to setup block scopes and pass it along. 

### Submitter checklist


jglick:
> I need access to the `FlowNodeMonitoringAction` to get the open telemetry trace parent id.

Sorry, I am not familiar with this is.

The diff looks reasonable, though as you noted it could be made binary compatible which might ease rollout.

timja:
not needed anymore

PR #625, opened 2024-07-09
Test history refactoring and improvements

mdealer:
This is an update of https://github.com/jenkinsci/junit-plugin/pull/446

Links to HPI iterations if you want to try it: 

https://github.com/mdealer/junit-plugin/releases/

TL;DR:
Major performance and usability improvements to the test history file storage, history charts and test failure age computation.

image

**Improve large JUnit test history performance and usability. RAM and CPU resources are better utilized to parse and cache the test results for faster retrieval.** 

JUnit SQL Storage plugin does not solve the problem fully as there are too many requests generated by the existing implementation.

To fully benefit from this, assign more CPU cores, e.g. 8+, and increase the Java heap size, e.g. -Xmx8g. 

### Changes:

- Test result XML Files (junitResult.xml) are now parsed using an XML reader instead of XStream.
- Test result history is loaded in parallel.
- Test result history window (start and end build number) is now dynamic, allowing better navigation to past results via the webpage.
- Test result history can now handle a much larger test and build counts in the thousands.
- Test history graphs have been reimplemented to avoid build trend plugin and using ECharts directly for greater flexbility.
- Test result age now scans back at most 25 builds without test results instead of resetting age at the first failed build
- Test trend graph is shown more often on build page (some issues still remain in Jenkins core)
- Some limited support for Dark Reader otherwise the graphs may turn unreadable 
- Added new parameter to junit pipeline step: keepTestNames for cases when your results already contain a namespace

Binary is now bigger due to a 3rd party math library for processing historical results.

The more CPU cores, the faster the result XML file parsing, the quicker the initial and subsequent history page load. The more heap size, the more test results will remain cached (using SoftReferences), the less CPU and Disk IO usage when browsing the results back and forth. 

### New parameter in 'junit' pipeline step: keepTestNames

Allow disabling the test name mangling when uploading results from multiple stages or parallel branches in a single build via 'keepTestNames' parameter. There was a change in the plugin a few years ago that basically broke the result URLs (e.g. if we are linking to the results from some external system), this is a way to undo this change. 

```
def c = {
    node(null) {
        writeFile file: 'result.xml', text: '<testsuite><testcase time="2.6875" classname="MyTest" name="Run 0" assertions="3 total, 0 failed, 3 succeeded"/></testsuite>'
        junit keepTestNames: true, testResults: 'result.xml'
    }
}

parallel 'p1': c, 'p2': c
```

### Related tickets that are potentially fixed

[...]

PR #630, opened 2024-07-29
Restore compatiblity with junit-sql-storage

timja:
and some minor cleanup at the same time, bunch of unused variables mostly.

Also fixed clicking on builds in the charts, variable wasn't defined properly before

cc @mdealer

noticed in https://github.com/jenkinsci/junit-sql-storage-plugin/pull/430
caused by https://github.com/jenkinsci/junit-plugin/pull/625

### Testing done

manually tested both junit and junit-sql-storage

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/History.java
+++ b/src/main/java/hudson/tasks/junit/History.java
@@ -23,23 +23,10 @@
  */
 package hudson.tasks.junit;
 
-import java.lang.ref.SoftReference;import java.math.RoundingMode;
-import java.text.DecimalFormat;import java.util.*;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.logging.Logger;
-import java.util.logging.Level;
-import java.util.stream.Collectors;
-
-import jenkins.util.SystemProperties;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
-import org.kohsuke.stapler.bind.JavaScriptMethod;
-
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.pivovarit.collectors.ParallelCollectors;
-
 import edu.hm.hafner.echarts.ChartModelConfiguration;
 import edu.hm.hafner.echarts.JacksonFacade;
 import edu.hm.hafner.echarts.LinesChartModel;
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/tasks/junit/History.java
```diff
@@ -170,30 +163,29 @@ private ObjectNode computeDurationTrendJson(List<HistoryTestResultSummary> histo
...
             domainAxisLabels.add(fdn);
-            ObjectNode durationColor = mapper.createObjectNode();
-            double duration = Math.round(mul * to.getDuration() * roundMul) / roundMul;
+            ObjectNode durationColor = MAPPER.createObjectNode();
+            double duration = Math.round(mul * h.getDuration() * roundMul) / roundMul;
             tmpMax = Math.max((float)duration, tmpMax);
-            lrY[index] = (double)(duration);
+            lrY[index] = duration;
             durationColor.put("value", duration);
-            if (to.isPassed() || (to.getPassCount() > 0 && to.getFailCount() == 0)) {
```
timja, line 191: minor behaviour change here but should be safe enough

src/main/java/hudson/tasks/junit/HistoryTestResultSummary.java
```diff
@@ -11,21 +13,19 @@ public class HistoryTestResultSummary {
     private final int skipCount;
     private final int passCount;
     private final String description;
-    private final hudson.tasks.test.TestResult resultInRun;
 
-    public HistoryTestResultSummary(Run<?, ?> run, hudson.tasks.test.TestResult resultInRun,
+    public HistoryTestResultSummary(Run<?, ?> run,
```
timja, line 17: this was the critical change, adding `resultInRun` broke the plugin and I can't see a way to retrieve it there. The API could be modified to maybe pass some more info down but it didn't seem necessary the only additional information that is missing now is the id of the test report which could be stored maybe if required

src/test/java/io/jenkins/plugins/junit/storage/TestResultStorageJunitTest.java
```diff
@@ -476,7 +476,7 @@ public List<HistoryTestResultSummary> getHistorySummary(int offset) {
                                     Job<?, ?> theJob = Jenkins.get().getItemByFullName(getJobName(), Job.class);
                                     if (theJob != null) {
                                         Run<?, ?> run = theJob.getBuildByNumber(buildNumber);
-                                        historyTestResultSummaries.add(new HistoryTestResultSummary(run, null, duration, failed, skipped, passed));
```
timja, line 479: this shouldn't have had null passed as it made the test run but if any one actually tried to access the history view they got null pointers

src/main/java/hudson/tasks/junit/HistoryTestResultSummary.java
mdealer: Any reason to not just pass null to it? Without the final URL it's very cumbersome to navigate single test or test subset history.

timja: can't remember offhand, would be best to keep the same features though i.e. it shouldn't be harder to navigate the tests when using pluggable storage

mdealer: Well, it is rather nearly impossible to navigate with 10K+ tests. This is what a user normally expects when viewing the history at a certain level. After all, the numbers shown are for the test and test set, not the overall build, but the URL is pointing to overall build, so it is a bug depending on the perspective.

timja: feel free to send a PR that maintains compatibility

PR #639, opened 2024-08-20
Remove dependencies on `commons-math3` and `ssj` by implementing simple linear regression directly, replacing the build duration distribution chart with a histogram, and deleting the smoothed trend lines in the test history charts

dwnusbaum:
This is a followup to https://github.com/jenkinsci/junit-plugin/pull/638 and https://github.com/jenkinsci/junit-plugin/pull/625 to further simplify the dependencies after I discussed it with some colleagues. This plugin is used pretty widely in the Jenkins ecosystem, so it seems worth trying to minimize its dependencies, especially to eliminate dependencies which are not being actively maintained.

This PR does 3 things (you can review commit by commit if you like):
* It gets rid of the `commons-math3` dependency by just directly implementing a simple linear regression algorithm. This is very straightforward since we are only using it for a trend line and don't care about any other data related to the regression.
* It replaces the `SmoothingCubicSpline`-based build duration distribution chart with a histogram, which I think is a better representation of the data anyway.
* It completely deletes the two `SmoothingCubicSpline`-based "Smooth of X" trend lines that are only shown in the test history widget if more than 200 builds' worth of test results are available. I did this so we could get rid of the `ssj` dependency completely. Here is my justification:
    * I do not think pluggable storage for test results is in widespread use, so most users set up build retention policies that will prevent them from getting anywhere close to 201 builds' worth of test results, so those users can never see these trend lines anyway
    * Even for users who do have at least 201 build's worth of test results, by default the page only shows 100 builds' worth test results, so only users who actively play around with the options on the page would ever see these trend lines
    * These trend lines were introduced for the first time in https://github.com/jenkinsci/junit-plugin/pull/625, so this is not removing long-lived functionality
    * I could not find any simple replacement for `SmoothingCubicSpline` in an actively-maintained library, and I am not confident in being able to reimplement it from scratch correctly in a reasonable amount of time
    * `ssj` is not actively maintained and has some warning signs, for example its 3.3.2 release is not listed anywhere in publicly available Git commit history
    * I am totally ok with adding this functionality back to the plugin with any of these three approaches:
        * If someone can find a replacement for `SmoothingCubicSpline` in an actively-maintained library with minimal dependencies and that ideally is not pulling in a whole grab bag of unrelated functionality that we do not need
        * If someone is willing to implement an equivalent algorithm directly
        * If someone wants to make it possible for https://github.com/jenkinsci/license-maven-plugin to support manually-specified license information, so that we could copy/paste the implementation of `SmoothingCubicSpline` into this repository while still having license information be reported accurately in Jenkins on the `/manage/about/` page

Here are screenshots with and without the PR:

With this PR:

Screen Shot 2024-08-20 at 1 46 41 PM

Without this PR:

Screen Shot 2024-08-20 at 1 45 38 PM

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/History.java
+++ b/src/main/java/hudson/tasks/junit/History.java
@@ -479,24 +479,16 @@ public class History {
     private ObjectNode computeDistributionJson(List<HistoryTestResultSummary> history) {
         ObjectNode root = MAPPER.createObjectNode();
         ArrayNode series = MAPPER.createArrayNode();
-        ArrayNode domainAxisLabels = MAPPER.createArrayNode();
 
         ObjectNode durationSeries = MAPPER.createObjectNode();
         durationSeries.put("name", "Build Count");
-        durationSeries.put("type", "line");
-        durationSeries.put("symbol", "circle");
-        durationSeries.put("symbolSize", "0");
-        durationSeries.put("sampling", "lttb");
+        durationSeries.put("type", "bar");
+        durationSeries.put("barWidth", "99%");
         ArrayNode durationData = MAPPER.createArrayNode();
         durationSeries.set("data", durationData);
         ObjectNode durationStyle = MAPPER.createObjectNode();
         durationSeries.set("itemStyle", durationStyle);
         durationStyle.put("color", "--success-color"); // "rgba(50, 200, 50, 0.8)");
-        durationSeries.put("stack", "stacked");
-        ObjectNode durAreaStyle = MAPPER.createObjectNode();
-        durationSeries.set("areaStyle", durAreaStyle);
-        durAreaStyle.put("color", "rgba(0,0,0,0)");
-        durationSeries.put("smooth", true);
         series.add(durationSeries);
 
         double maxDuration = 0, minDuration = Double.MAX_VALUE;
@@ -508,22 +500,20 @@ public class History {
                 minDuration = h.getDuration();
             }
         }
-        double extraDuration = Math.max(0.001, (maxDuration - minDuration) * 0.05);
-        minDuration = Math.max(0.0, minDuration - extraDuration);
-        maxDuration = maxDuration + extraDuration;
-        int[] counts = new int[100];
-        int smoothBuffer = 2;
-        double[] lrX = new double[counts.length + smoothBuffer * 2 + 1];
-        double[] lrY = new double[counts.length + smoothBuffer * 2 + 1];
-        double scale = maxDuration - minDuration;
-        double step = scale / counts.length;
+        // double extraDuration = Math.max(0.001, (maxDuration - minDuration) * 0.05);
+        minDuration = Math.max(0.0, minDuration);
+        int buckets = 50;
+        double[] lrX = new double[buckets];
+        double[] lrY = new double[buckets];
+        double domain = maxDuration - minDuration;
+        double step = domain / buckets;
         for (HistoryTestResultSummary h : history) {
-            int idx = smoothBuffer + (int) Math.round((h.getDuration() - minDuration) / step);
+            int idx = (int) Math.round((h.getDuration() - minDuration) / step);
             int idx2 = Math.max(0, Math.min(idx, lrY.length - 1));
             lrY[idx2]++;
         }
         for (int i = 0; i < lrY.length; ++i) {
-            lrX[i] = ((minDuration + (maxDuration - minDuration) / lrY.length * i) / scale * 100.0);
+            lrX[i] = minDuration + step * (i + 0.5);
         }
 
         ObjectNode xAxis = MAPPER.createObjectNode();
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/tasks/junit/History.java
```diff
@@ -483,24 +421,16 @@ private ObjectNode computeResultTrendJson(List<HistoryTestResultSummary> history
...
-        ArrayNode domainAxisLabels = MAPPER.createArrayNode();
 
         ObjectNode durationSeries = MAPPER.createObjectNode();
         durationSeries.put("name", "Build Count");
-        durationSeries.put("type", "line");
-        durationSeries.put("symbol", "circle");
-        durationSeries.put("symbolSize", "0");
-        durationSeries.put("sampling", "lttb");
+        durationSeries.put("type", "bar");
+        durationSeries.put("barWidth", "99%");
```
dwnusbaum, line 428: Slightly less than 100% so you see a border between each bar.

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -483,24 +421,16 @@ private ObjectNode computeResultTrendJson(List<HistoryTestResultSummary> history
...
         ArrayNode durationData = MAPPER.createArrayNode();
         durationSeries.set("data", durationData);
         ObjectNode durationStyle = MAPPER.createObjectNode();
         durationSeries.set("itemStyle", durationStyle);
         durationStyle.put("color", "--success-color"); // "rgba(50, 200, 50, 0.8)");
```
dwnusbaum, line 433: This color is a bit ugly now that it is not just a simple line in case anyone has any recommendation. Ideally this would be a stacked histogram chart with each build result shown in a different color.

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -512,22 +442,20 @@ private ObjectNode computeDistributionJson(List<HistoryTestResultSummary> histor
...
-        double extraDuration = Math.max(0.001, (maxDuration - minDuration) * 0.05);
-        minDuration = Math.max(0.0, minDuration - extraDuration);
-        maxDuration = maxDuration + extraDuration;
-        int[] counts = new int[100];
-        int smoothBuffer = 2;
-        double[] lrX = new double[counts.length + smoothBuffer * 2 + 1];
-        double[] lrY = new double[counts.length + smoothBuffer * 2 + 1];
-        double scale = maxDuration - minDuration;
-        double step = scale / counts.length;
+        // double extraDuration = Math.max(0.001, (maxDuration - minDuration) * 0.05);
```
dwnusbaum, line 445: Remove commented-out line

```suggestion
```

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -512,22 +442,20 @@ private ObjectNode computeDistributionJson(List<HistoryTestResultSummary> histor
...
+        double step = domain / buckets;
         for (HistoryTestResultSummary h : history) {
-            int idx = smoothBuffer + (int) Math.round((h.getDuration() - minDuration) / step);
+            int idx = (int) Math.round((h.getDuration() - minDuration) / step);
             int idx2 = Math.max(0, Math.min(idx, lrY.length - 1));
             lrY[idx2]++;
         }
         for (int i = 0; i < lrY.length; ++i) {
-            lrX[i] = ((minDuration + (maxDuration - minDuration) / lrY.length * i) / scale * 100.0);
+            lrX[i] = minDuration + step * (i + 0.5);
```
dwnusbaum, line 458: Offset the X values so they are in the middle of the bucket. We are no longer scaling them into the range 0-1 because we are using them directly in our data series.

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -551,24 +479,25 @@ private ObjectNode computeDistributionJson(List<HistoryTestResultSummary> histor
...
-        double k = (double) counts.length / smoothPts;
-        final double splineRoundMul = 1000.0;
-        for (double z = minDuration; z < maxDuration; z += step * k) {
-            double v = Math.round(splineRoundMul * Math.max(0.0, scs.evaluate(z / scale * 100.0))) / splineRoundMul;
-            durationData.add((float) v);
-            maxBuilds = Math.max(maxBuilds, (int) Math.ceil(v));
-            // Use float for smaller JSONs.
-            domainAxisLabels.add((float) (Math.round(mul * z * roundMul) / roundMul));
+        for (int i = 0; i < lrY.length; i++) {
+            double scaledX = mul * lrX[i];
```
dwnusbaum, line 489: We no longer round our X values in the backend (`z` in the old code), because our axis type is now `value` instead of `category`, and so if we round here our bars will have weird gaps and not be contiguous.

src/main/java/hudson/tasks/junit/History.java
```diff
@@ -551,24 +479,25 @@ private ObjectNode computeDistributionJson(List<HistoryTestResultSummary> histor
...
+            double y = lrY[i];
```
dwnusbaum, line 490: Nothing fancy with the Y values anymore, this is just the frequency of build durations in this bucket. We can probably make `lrY` an `int[]` now.

src/main/resources/hudson/tasks/junit/History/history.js
```diff
@@ -323,12 +323,16 @@ function onBuildIntervalChange(selectObj) {
...
                 },
                 xAxis: {
-                    type: 'category',
-                    boundaryGap: false,
+                    type: 'value',
                     axisLabel: {
-                        color: textColor
+                        color: textColor,
+                        formatter: function(value) {
+                            return Math.round(value * model.distribution.xAxis.roundingFactor) / model.distribution.xAxis.roundingFactor;
```
dwnusbaum, line 330: Round the labels in the frontend.

src/main/resources/hudson/tasks/junit/History/history.js
```diff
@@ -323,12 +323,16 @@ function onBuildIntervalChange(selectObj) {
...
+                        }
                     },
-                    data: model.distribution.domainAxisLabels,
+                    min: model.distribution.xAxis.min,
+                    max: model.distribution.xAxis.max,
+                    minInterval: model.distribution.xAxis.interval,
```
dwnusbaum, line 335: What I really wanted was to be able to tell `echarts` to pick an interval that is a multiple of `model.distribution.xAxis.interval`, and then let it automatically choose which multiple based on screen and label size, but I didn't see any way to do it.

If you set `interval: model.distribution.xAxis.interval`, then you get one grid line per bucket, which is nice, but the labels are unreadably close to each other and I couldn't find a way to fix that dynamically. You can set `interval: model.distribution.xAxis.interval * 2` and things look fine on "normalish" laptop screen size, but again it doesn't scale at all if you shrink your screen size.

I also wanted to have dynamic `minorTick` support, e.g. if we let `echarts` pick the interval, could we at least set up the minor ticks to align with the buckets? Unfortunately I don't think so, you can only control how many minor ticks you want between each major tick, e.g. if you specify `interval: model.distribution.xAxis.interval * 3`, you could specify `minorTick { show: true, splitNumber: 3 }`, but again that doesn't adjust based on the data or screen size at all.

timja (approved):
Thanks for the PR!

I've tested with 500 builds each with roughly 8,000 test cases and performance was good. Results look similar to before except for in the distribution graph which has changed

PR #640, opened 2024-08-22
Improve recently added log messages by adding context to indicate which build is relevant

dwnusbaum:
Amends #625. @olamy recently saw a bunch of these log messages, which were added in that PR I guess to help detect slow loading, but they are not really useful without indicating which build is relevant:

```
2024-08-22 08:03:49.329+0000 [id=465]	WARNING	h.tasks.junit.TestResultAction#getResult: TestResultAction.load took 1060 ms.
```

This PR does 3 main things:
* For the new log messages added in #625, it adds either `Run.toString()` or `File.toString()` to the message to provide context
* It removes the hard-coded `Class.method` parts of the messages, which are added automatically by the default Jenkins logger
* It changes the visibility of the new static `parse` methods in #625 that as far as I understand were only intended for internal use from `public` to package-private. This plugin is a dependency of various other plugns, so I think it is desirable to avoid accidentally increasing the API surface

### Testing done

I mangled some test files locally and enabled FINEST logging to check that log messages worked and indicated which build was relevant.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -41,7 +41,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
@@ -269,7 +268,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         return Math.min(365.0f * 24 * 60 * 60, Math.max(0.0f, d));
     }
 
-    public static CaseResult parse(SuiteResult parent, final XMLStreamReader reader, String ver)
+    static CaseResult parse(SuiteResult parent, final XMLStreamReader reader, String context, String ver)
             throws XMLStreamException {
         CaseResult r = new CaseResult(parent, null, null, null);
         while (reader.hasNext()) {
@@ -318,19 +317,17 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
                         break;
                     case "properties":
                         r.properties = new HashMap<>();
-                        parseProperties(r.properties, reader, ver);
+                        parseProperties(r.properties, reader, context, ver);
                         break;
                     default:
-                        if (LOGGER.isLoggable(Level.FINEST)) {
-                            LOGGER.finest("CaseResult.parse encountered an unknown field: " + elementName);
-                        }
+                        LOGGER.finest(() -> "Unknown field in " + context + ": " + elementName);
                 }
             }
         }
         return r;
     }
 
-    public static void parseProperties(Map<String, String> r, final XMLStreamReader reader, String ver)
+    static void parseProperties(Map<String, String> r, final XMLStreamReader reader, String context, String ver)
             throws XMLStreamException {
         while (reader.hasNext()) {
             final int event = reader.next();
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -269,7 +268,7 @@ public static float clampDuration(float d) {
         return Math.min(365.0f * 24 * 60 * 60, Math.max(0.0f, d));
     }
 
-    public static CaseResult parse(SuiteResult parent, final XMLStreamReader reader, String ver)
```
dwnusbaum, line 272: As far as I know nothing should be using these new methods from #625, so I made them all package-private, but if that is not true, let me know and I will revert that part of the change.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -318,19 +317,17 @@ public static CaseResult parse(SuiteResult parent, final XMLStreamReader reader,
...
                     case "properties":
                         r.properties = new HashMap<>();
-                        parseProperties(r.properties, reader, ver);
+                        parseProperties(r.properties, reader, context, ver);
                         break;
                     default:
-                        if (LOGGER.isLoggable(Level.FINEST)) {
-                            LOGGER.finest("CaseResult.parse encountered an unknown field: " + elementName);
-                        }
+                        LOGGER.finest(() -> "Unknown field in " + context + ": " + elementName);
```
dwnusbaum, line 323: I also checked whether `XMLReader.getLocation` included a file name to avoid having to pass `context` everywhere, but `getPublicID` and `getSystemID` just return null. We could add line and column numbers to give even more context, but I don't really think it would be helpful in practice.

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -179,8 +179,8 @@ public synchronized TestResult getResult() {
             skipCount = r.getSkipCount();
         }
         long d = System.nanoTime() - started;
-        if (d > 500000000L && LOGGER.isLoggable(Level.WARNING)) {
```
dwnusbaum, line 182: 500000000 is 5e8, or 0.5 seconds. I adjusted the code to use `TimeUnit` to make it clearer, but I could switch it to 500 * 1000 * 1000 or whatever you think is clearest.

src/main/java/hudson/tasks/junit/TestResultAction.java
```diff
@@ -69,7 +70,6 @@
 @SuppressFBWarnings(value = "UG_SYNC_SET_UNSYNC_GET", justification = "False positive")
 public class TestResultAction extends AbstractTestResultAction<TestResultAction>
         implements StaplerProxy, SimpleBuildStep.LastBuildAction {
-    private static final Logger LOGGER = Logger.getLogger(TestResultAction.class.getName());
```
dwnusbaum, line 72: This was added in #625, but we already had a logger at the bottom of the file, it was just named lowercase `logger`.

src/main/java/hudson/tasks/junit/TestResultAction.java
dwnusbaum: (I also wonder if this threshold is too low for `WARNING` and it should be increased to 1-2 seconds, maybe logging at `FINE` for stuff between 0.5-1 seconds)

MarkEWaite: I like the idea of setting the threshold higher for logging at less than 2 seconds, but approve this pull request with or without that change.

MarkEWaite (approved):

timja (approved):

olamy:
@dwnusbaum Thanks!

PR #451, opened 2022-10-24
[JENKINS-69658] - failed-test.jelly javascript un-inlined.

Artmorse:
**The issue**
You can find all the issue details [here](https://issues.jenkins.io/browse/JENKINS-69658).

**My updates**
I've moved the `onclick` calls from the `failed-test.jelly` file to the `failureSummary.js` one.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #654, opened 2024-10-06
 [JENKINS-69658] - CSP compatibility for summary.jelly

shlomomdahan:
[JENKINS-69658](https://issues.jenkins.io/browse/JENKINS-69658)

This PR is an extension of https://github.com/jenkinsci/junit-plugin/pull/451 to fix the inline JS in the summary.jelly file.

The failureSummary.js file had to be changed so that it can handle event delegation as some of the elements we are targeting with the fix do not actually appear in the html until the parent is expanded. 

### Testing done

- STATUS QUO: No changes implemented. This is the intended functionality. 

https://www.loom.com/share/796f46d313e34c8cbbaf85963c1ffb59?sid=20dc0f68-ddca-4f88-a977-1580816dcdab

- NON RESTRICTIVE: This is the behavior after implementing the CSP changes. The code is working just as before and there is no regression in functionality.

https://www.loom.com/share/f7a8117e45a54da5a719599f0d1adb08?sid=8c812aef-9c9a-4a81-b662-c49229d68dbf

- RESTRICTIVE: This is with CSP restrictive mode enabled. It is working correctly as intended without any issues. 

https://www.loom.com/share/6e3642baf5934fe59e2efdc3f3f0ec3b

### Submitter checklist


src/main/resources/lib/hudson/test/js/failureSummary.js
```diff
@@ -24,24 +27,30 @@ function hideFailureSummary(summaryId) {
...
-    const testShowlinks = document.querySelectorAll("a[id*=test-][id*=-showlink]");
-    testShowlinks.forEach((element) => {
-        element.addEventListener('click', (event) => {
-            const id = element.id.replace(PREFIX, '').replace(SHOWLINK_SUFFIX, '');
-            const summaryId = PREFIX + id;
-            showFailureSummary(summaryId, document.URL + id + "summary");
-        })
+    container.querySelectorAll('a[id$="-showlink"], a[id$="-hidelink"]').forEach(link => {
+        if (!link.onclick) {
+            link.onclick = handleShowHideClick;
```
timja, line 35: please use `addEventListener` like the code removed was using, see https://stackoverflow.com/questions/6348494/addeventlistener-vs-onclick

shlomomdahan: Sure - I just made this correction in the latest commit.

timja (approved):
Thanks looks good, (untested)

src/main/resources/lib/hudson/test/js/failureSummary.js
daniel-beck: (Disclaimer: Still haven't run this.) This change request looks wrong since unlike the earlier code, the events are re-initialized on HTTP response without removing previous event listeners? What happens after a few `showFailureSummary`, does that collect event listeners?

shlomomdahan: Hi @daniel-beck 

If I understand correctly, this shouldn't be an issue with the new change:

```
container.querySelectorAll('a[id$="-showlink"], a[id$="-hidelink"]').forEach(link => {
        link.addEventListener('click', handleShowHideClick);
        link.style.cursor = 'pointer';
```

since duplicate event listeners cannot be added to elements [see docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#:~:text=If%20the%20function%20or%20object%20is%20already%20in%20the%20list%20of%20event%20listeners%20for%20this%20target%2C%20the%20function%20or%20object%20is%20not%20added%20a%20second%20time.)

https://stackoverflow.com/questions/10364298/what-does-calling-addeventlister-twice-do

Is this what you were referring to?

Thanks

daniel-beck: Ah, good point, I had only looked at the previous reference. So this won't cause trouble then. Thanks for the link!

PR #664, opened 2024-12-01
URL encode fix for All Failed Tests sections

ahadalioglu:
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #667, opened 2024-12-10
Brackets fix for all failed tests sections

ahadalioglu:
### Testing done through Jenkins: [https://ci.jenkins.io/job/Plugins/job/junit-plugin/job/PR-664/](https://ci.jenkins.io/job/Plugins/job/junit-plugin/job/PR-664/)

After the fix implemented in [PR #661](https://github.com/jenkinsci/junit-plugin/pull/661), the "All Failed Tests" sections continued to fail due to the lack of percent-encoding for square brackets specifically.

This PR addresses and resolves the issue.

### Submitter checklist

### DIFF
--- a/src/main/resources/hudson/tasks/test/MetaTabulatedResult/body.jelly
+++ b/src/main/resources/hudson/tasks/test/MetaTabulatedResult/body.jelly
@@ -40,14 +40,15 @@ THE SOFTWARE.
       </thead>
       <j:forEach var="f" items="${it.failedTests}" varStatus="i">
         <tr>
-          <td class="pane no-wrap">
-            <t:failed-test result="${f}">
-              <j:invokeStatic var="encodedRelativePath" className="hudson.Util" method="rawEncode">
-                <j:arg value="${f.getRelativePathFrom(it)}" />
-              </j:invokeStatic>
-              <a href="${encodedRelativePath}">${f.name}</a>
-            </t:failed-test>
-          </td>
+          <!-- Temporarily replace slashes to prevent rawEncode from encoding them. -->
+          <j:set var="tempPath" value="${f.getRelativePathFrom(it).replace('/', '__SLASH__')}" />
+          <j:set var="rawEncodedPath">
+            <j:invokeStatic className="hudson.Util" method="rawEncode">
+              <j:arg value="${tempPath}" />
+            </j:invokeStatic>
+          </j:set>
+          <j:set var="encodedRelativePath" value="${rawEncodedPath.replace('__SLASH__', '/')}" />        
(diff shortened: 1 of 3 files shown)


basil (changes requested):
Inconsistent with #661—should call `Util#rawEncode` on each component of the path internally within `getRelativePathFrom` as suggested in https://github.com/jenkinsci/junit-plugin/pull/664#issuecomment-2515723693.

ahadalioglu:
Although we used `rawEncode` with `p.safeName` and it worked, `rawEncode` unfortunately encodes the entire string, including slashes (/) as %2F. This becomes problematic when dealing with paths that require the slash delimiter to separate components.

To address this, I modified the `hudson.Util` class by adding a method called `encodeEachPathComponent`. This method splits the path into components using the slash (/) as a delimiter and applies encoding to each component individually.

As a result, the encoded URL now preserves the slashes while ensuring that each component is safely encoded.

Please review it at your earliest convenience.

basil:
> To address this, I modified the `hudson.Util` class by adding a method called `encodeEachPathComponent`.

Where is the code for that method?

ahadalioglu:
Oh, sorry, I have missed that:
```
public static String encodeEachPathComponent(String path) {
    String[] components = path.split("/");  // Split by "/"
    StringBuilder encodedPath = new StringBuilder();
    
    for (int i = 0; i < components.length; i++) {
        if (i > 0) encodedPath.append("/");  // Re-add the slash between components
        encodedPath.append(URLEncoder.encode(components[i], StandardCharsets.UTF_8));  // Encode each component
    }

    return encodedPath.toString();
}
```

ahadalioglu:
It is done. Is it good now?

ahadalioglu:
Now, it's ready with my above-mentioned changes. 

Additionally, I have another solution: 
Based on our records, the only issue with the `All Failed Tests` links is the brackets. If we replace just those two characters ([ and ]) with %5B and %5D, there's no need for further encoding (such as using Util/rawEncode), as there are no other issues with the broken links:
```
<j:set var="rawPath" value="${f.getRelativePathFrom(it)}" />
<j:set var="encodedRelativePath">
    ${rawPath.replace('[', '%5B').replace(']', '%5D')}
</j:set>
```

Please let me know your thoughts.

basil:
As stated in https://github.com/jenkinsci/junit-plugin/pull/667#pullrequestreview-2496688087 the solution should be consistent with #661 by encoding any characters that need to be encoded, not just square brackets. To accomplish this, the `getRelativePathFrom` code in `src/main/java/hudson/tasks/test/TestObject.java` needs to be updated. The call to `cur.getSafeName` probably needs to be wrapped in `Util#rawEncode`. In addition, the change needs to be tested with a test case containing square brackets.

PR #701, opened 2025-05-11
Prevent stdio trimming from creating invalid unicode data

rmuir:
`CaseResult.possiblyTrimStdio()` creates substrings according to UTF-16 (java `char`) boundaries. When stdio contains Unicode supplementary characters, this method can create unpaired surrogates by splitting a single unicode character "in half". Such strings can't be converted to UTF-8 and cause build failures:

```
FATAL: Failed to save the JUnit test result
java.io.IOException: java.lang.RuntimeException: Failed to serialize hudson.tasks.junit.TestResult#suites for class hudson.tasks.junit.TestResult
Caused by: java.lang.RuntimeException: Failed to serialize hudson.tasks.junit.TestResult#suites for class hudson.tasks.junit.TestResult
Caused by: java.lang.RuntimeException: Failed to serialize hudson.tasks.junit.SuiteResult#stderr for class hudson.tasks.junit.SuiteResult
Caused by: [com.thoughtworks.xstream.io](http://com.thoughtworks.xstream.io/).StreamException: Invalid character 0xd83a in XML stream
```

Fix trimming to remove unpaired trailing (low) surrogates at the beginning of truncated strings and unpaired leading (high) surrogates at the end of truncated strings.

We're hitting this bug over at Apache Jenkins every once in a while: in the Lucene project our tests exercise a lot of Unicode functionality due to the nature of the software.

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -375,6 +375,23 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         }
     }
 
+    /**
+     * Cleans up a truncated string, that might contain unpaired surrogates
+     * @param str input string
+     * @return input string, possibly truncated to avoid unpaired surrogates
+     */
+    static CharSequence cleanupTruncated(CharSequence str) {
+        // remove unpaired trailing surrogates at the start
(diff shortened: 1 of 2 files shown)


timja (approved):
Thanks!

Issue #704, opened 2025-05-13
build's full "Test Results" page is displayed under a single test case


car-roll:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>
  
Jenkins: 2.479.3
OS: Mac OS X - 15.4.1
Java: 23.0.2 - Homebrew (OpenJDK 64-Bit Server VM)
---
antisamy-markup-formatter:173.v680e3a_b_69ff3
apache-httpcomponents-client-4-api:4.5.14-269.vfa_2321039a_83
asm-api:9.8-135.vb_2239d08ee90
bootstrap5-api:5.3.3-2
bouncycastle-api:2.30.1.80-256.vf98926042a_9b_
branch-api:2.1217.v43d8b_b_d8b_2c7
byte-buddy-api:1.17.5-147.v9ce9318b_4ce8
caffeine-api:3.2.0-166.v72a_6d74b_870f
checks-api:370.vb_61a_c57328f3
cloudbees-folder:6.1012.v79a_86a_1ea_c1f
commons-compress-api:1.26.1-2
commons-lang3-api:3.17.0-87.v5cf526e63b_8b_
commons-text-api:1.13.0-153.v91dcd89e2a_22
configuration-as-code:1963.v24e046127a_3f
credentials:1415.v831096eb_5534
credentials-binding:687.v619cb_15e923f
database:274.vea_2e859b_2661
database-h2:1.4.116.vc979ffa_d29c7
display-url-api:2.209.v582ed814ff2f
durable-task:587.v84b_877235b_45
echarts-api:5.6.0-4
font-awesome-api:6.7.2-1
git:5.7.0
git-client:6.1.3
github:1.43.0
github-api:1.321-488.v9b_c0da_9533f8
github-branch-source:1815.v9152b_2ff7a_1b_
github-checks:581.va_9669c91d6cb_
gson-api:2.13.1-139.v4569c2ef303f
h2-api:11.1.4.199-36.vb_ee07e965744
instance-identity:203.v15e81a_1b_7a_38
ionicons-api:88.va_4187cb_eddf1
jackson2-api:2.18.3-402.v74c4eb_f122b_2
jakarta-activation-api:2.1.3-2
jakarta-mail-api:2.1.3-2
javax-activation-api:1.2.0-8
jaxb:2.3.9-133.vb_ec76a_73f706
jjwt-api:0.11.5-120.v0268cf544b_89
[...]

PR #705, opened 2025-05-13
Fix expand single failure

car-roll:
Addresses https://github.com/jenkinsci/junit-plugin/issues/704
Removes any search parameter from the URL when showing test failures

### Testing done
Manual testing. Created job with failure report. Visited the test results page with a search query and then clicked on the failures.

### Submitter checklist

### DIFF
--- a/src/main/resources/lib/hudson/test/js/failureSummary.js
+++ b/src/main/resources/lib/hudson/test/js/failureSummary.js
@@ -43,7 +43,9 @@ function handleShowHideClick(event) {
     const id = link.id.replace(/-showlink$/, '').replace(/-hidelink$/, '');
 
     if (link.id.endsWith('-showlink')) {
-        showFailureSummary(id, document.URL + id.replace(PREFIX, '') + "summary");
+        // Remove any queries from URL
+        const baseUrl = document.URL.split('?')[0];
+        showFailureSummary(id, baseUrl + id.replace(PREFIX, '') + "summary");
     } else {
         hideFailureSummary(id);
     }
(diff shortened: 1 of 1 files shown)


src/main/resources/lib/hudson/test/js/failureSummary.js
```diff
@@ -43,7 +43,10 @@ function handleShowHideClick(event) {
     const id = link.id.replace(/-showlink$/, '').replace(/-hidelink$/, '');
 
     if (link.id.endsWith('-showlink')) {
-        showFailureSummary(id, document.URL + id.replace(PREFIX, '') + "summary");
+        // clear the query parameters
+        const cleanUrl = new URL(document.URL);
+        cleanUrl.search = "";
```
timja, line 48: simpler:
```suggestion
        const cleanUrl = location.toString().replace(location.search, "");
```

jglick: :shrug: Setting `.search` directly seems clearer and more reliable to me.

timja: It’s already clear enough that it needs a comment explaining it but sure we can go with this

timja:
> Something still seems fishy with using string concatenation to append to the path component of the current page’s URL. Does JavaScript not have an equivalent to `java.net.URI.resolve`?

Not that I can tell, the only appending is for adding query parameters which is common.

Yeah relative path should fix it.

jtnord:
> > Something still seems fishy with using string concatenation to append to the path component of the current page’s URL. Does JavaScript not have an equivalent to `java.net.URI.resolve`?
> 
> Not that I can tell, the only appending is for adding query parameters which is common.

`URL(relativeOrAbsulute, baseUrl)` -> https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references?

PR #730, opened 2025-09-03
Disable automatic test file validation

tiago-espinha-img:
Automatic test file validation causes issues when workspaces have large O(millions) amount of files. In those cases, if the particular test file doesn't exist at the time the user looks at the configuration page, the Agent will start crawling the whole workspace looking for other subdirectories where that file might be.

By doing so, it's tracking all the files it's seen which eventually leads to filling up the heap and crashing.

This disables the automatic file validation.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
+++ b/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java
@@ -35,7 +35,6 @@ import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
 import hudson.model.Descriptor;
-import hudson.model.Item;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.Saveable;
@@ -59,7 +58,6 @@ import java.util.logging.Logger;
 import jenkins.tasks.SimpleBuildStep;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.types.FileSet;
-import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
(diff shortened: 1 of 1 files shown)


tiago-espinha-img:
FWIW, usability regression or not, this validation seems entirely useless AND dangerous. Why would anyone want to validate that the artifacts produced by a job are present at _configuration-time_?

And sure, the defaults seem unfit for purpose, but how small is too small? How long is a piece of string?

It's a complete ticking bomb that the UI for _configuring_ the job can bring down an Agent, and all down to how much heap that Agent is configured with + the size of the workspace.

To put it in perspective, this has cost us many days worth of lost work due to Agents mysteriously falling over. When this happens, the symptoms are completely generic. The heap starts filling up and eventually the JVM runs out of heap and the JNLP Agent falls over.

All because someone happened to open the job configuration page when the test artifacts weren't present... A completely innocent action, and impossible to correlate cause -> effect. In the end it took taking a heap dump at just the right time and painstakingly going through it with a fine comb to work out what was causing the large heap size.

I was trying to prevent others from hitting the same pain by completely disabling this. Tweaking the defaults might help, but again, how small is too small? At what point would we argue that too small a defaults would also constitute a "usability regression"?

As a final note, for anyone else facing this, we've come up with a workaround for it (intercepting the `^check*` HTTP requests, and responding with an empty HTTP OK, but this does require having a reverse proxy in place). It's not pretty but it works... and this way we're shielding ourselves from what IMO is poor default behavior.

jglick:
> how small is too small?

Clearly much smaller than it currently is.

tiago-espinha-img:
> Could you try set this system property, or in the script console: https://github.com/jenkinsci/jenkins/blob/cadd17a7fec0eb73391b8720fefe32db3123bc2f/core/src/main/java/hudson/FilePath.java#L3144
> 
> ```
> hudson.FilePath.VALIDATE_ANT_FILE_MASK_BOUND=100
> ```

I've tried this and it doesn't make a difference.

Even with it set to as little as 1:
image

These checks are taking O(minutes) with around 400k files in the workspace (this is a smaller sample I have as an attempt to reproduce the actual issue).

jglick:
Not sure offhand, but whatever is wrong, it should be fixed in core (at worst by deprecating this method).

PR #732, opened 2025-09-08
race condition and missing results in aggregation

jtnord:
In attempting to diagnose an ATH failure I noticed what looks like a potential race condition that could lead to results not being correctly updated.

There is also issue where if jobs are not specified and are to be inferre,d then these are obtained from the dependency graph at action creation time, but the graph may not be correct at this point in time as it is built asynchronously.

There is also an other timing issue.
For example in rare cases (but maybe not so rare in our ATH) if 2 builds finish at T1, a web request comes in a T2 and then another build fishes at T3 followed by a web request at T4.

When T1 arrives the timestamp is set to T1, and then in T2 the current value is checked against the timestamp and set to T1+1

When T3 arrives the timestamp is set to T3.

THe issue occurs if you have multiple CPUs and updating the code is fast and T1 = T2 = T3.

When the request at T4 arrives the code thinks everything is up-to-date when it is not.

### Testing done

internal ATH consistently failed before these changes, and passes after.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/test/AggregatedTestResultPublisher.java
+++ b/src/main/java/hudson/tasks/test/AggregatedTestResultPublisher.java
@@ -40,7 +40,6 @@ import hudson.model.ItemGroup;
 import hudson.model.Job;
 import hudson.model.Result;
 import hudson.model.Run;
-import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
@@ -352,7 +351,7 @@ public class AggregatedTestResultPublisher extends Recorder {
         public static class RunListenerImpl extends RunListener<Run> {
             @SuppressFBWarnings(value = "ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD", justification = "intended by design")
             @Override
-            public void onCompleted(Run run, TaskListener listener) {
+            public void onFinalized(Run run) {
                 lastChangedReference = new Object();
             }
         }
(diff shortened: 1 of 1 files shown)


jtnord:
Note, this did not help the failing ATH, but I think it is still a valid robustness fix.

timja (approved):

src/main/java/hudson/tasks/test/AggregatedTestResultPublisher.java
```diff
@@ -126,13 +127,9 @@ public static final class TestResultAction extends AbstractTestResultAction {
...
         /**
-         * The last time the fields of this object is computed from the rest.
+         * flag set when we need to recompute the aggregation.
          */
-        private transient long lastUpdated = 0;
-        /**
-         * When was the last time any build completed?
-         */
-        private static long lastChanged = 0;
+        private static final AtomicBoolean RECOMPUTE_NEEDED = new AtomicBoolean(true);
```
jtnord, line 132: I don't think this works correctly as there will be an action in multiple projects.  and all of them need to recompute, not just the first one.

src/main/java/hudson/tasks/test/AggregatedTestResultPublisher.java
```diff
@@ -144,43 +145,41 @@ public static final class TestResultAction extends AbstractTestResultAction {
...
 
         /**
          * Gets the jobs to be monitored.
          */
         public Collection<AbstractProject> getJobs() {
+            if (jobs == null) {
+                Set<AbstractProject> projects = getProject().getTransitiveDownstreamProjects();
+                return projects.stream()
+                        .filter(p -> p.hasPermission(Item.READ))
+                        .collect(Collectors.toSet());
```
jtnord, line 169: need to keep the spotless overlord hapy
```suggestion
                return projects.stream().filter(p -> p.hasPermission(Item.READ)).collect(Collectors.toSet());
```

timja:
> @timja I believe this is ready to merge (I have no karma to merge)

O I assumed you did.

Issue #747, opened 2025-10-14
Age link is broken


timja:
### Jenkins and plugins versions report

.

### What Operating System are you using (both controller, and any agents involved in the problem)?

.

### Reproduction steps

https://ci.jenkins.io/job/Core/job/jenkins/job/master/lastUnstableBuild/testReport/

<img width="1540" height="147" alt="Image" src="https://github.com/user-attachments/assets/18609486-6ea1-439b-b08a-f96b5a3b9ca4" />

### Expected Results

not sure where it should go but not there

### Actual Results

The `1` goes to https://ci.jenkins.io

PR #752, opened 2025-10-14
Don't show broken link when test is passing or skipped

timja:
Fixes https://github.com/jenkinsci/junit-plugin/issues/747

Alternative if people prefer:

```diff
diff --git i/src/main/java/hudson/tasks/junit/CaseResult.java w/src/main/java/hudson/tasks/junit/CaseResult.java
index fb09c1ac..be4499c9 100644
--- i/src/main/java/hudson/tasks/junit/CaseResult.java
+++ w/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -715,7 +715,7 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
      */
     @Exported(visibility = 9)
     public int getAge() {
-        if (isPassed()) {
+        if (isPassed() || isSkipped()) {
             return 0;
         } else if (getRun() != null) {
             return getRun().getNumber() - getFailedSince() + 1;
diff --git i/src/main/resources/lib/hudson/test/table.jelly w/src/main/resources/lib/hudson/test/table.jelly
index 3841a796..7b74f0b4 100644
--- i/src/main/resources/lib/hudson/test/table.jelly
+++ w/src/main/resources/lib/hudson/test/table.jelly
@@ -84,7 +84,7 @@ THE SOFTWARE.
                 </td>
                 <td class="jenkins-mobile-hide">
                     <j:choose>
-                        <j:when test="${f.age == 0 or f.skipped}">0</j:when>
+                        <j:when test="${f.age == 0}">${f.age}</j:when>
                         <j:otherwise>
                             <a href="${rootURL}/${f.failedSinceRun.url}">${f.age}</a>
                         </j:otherwise>
```

It would change the `getAge` API to not show skipped in age.
I think its not that useful unless e.g. you conditionally skip based on some code and want to see when that started happening maybe.

----

`failedSince` and `isPassed` are out of sync.

`isPassed` says skipped is not a pass.
`failedSince` does not consider skipped a failure.

Which is what causes this bug in the case of skipped.

----

Another alternative would be for `failedSince` to consider skipped as well

### Testing done

Tested with:
* skipped
* passed
* failure
* failing for multiple builds

### Submitter checklist

### DIFF
--- a/src/main/resources/lib/hudson/test/table.jelly
+++ b/src/main/resources/lib/hudson/test/table.jelly
@@ -83,7 +83,12 @@ THE SOFTWARE.
                     </j:forEach>
                 </td>
                 <td class="jenkins-mobile-hide">
-                    <a href="${rootURL}/${f.failedSinceRun.url}">${f.age}</a>
+                    <j:choose>
+                        <j:when test="${f.age == 0 or f.skipped}">0</j:when>
+                        <j:otherwise>
+                            <a href="${rootURL}/${f.failedSinceRun.url}">${f.age}</a>
+                        </j:otherwise>
+                    </j:choose>
                 </td>
                 <td class="no-wrap jenkins-mobile-hide" data="${f.duration}">
                     ${f.durationString}


src/main/resources/lib/hudson/test/table.jelly
```diff
@@ -83,7 +83,12 @@ THE SOFTWARE.
                     </j:forEach>
                 </td>
                 <td class="jenkins-mobile-hide">
-                    <a href="${rootURL}/${f.failedSinceRun.url}">${f.age}</a>
+                    <j:choose>
+                        <j:when test="${f.age == 0 or f.skipped}">0</j:when>
```
timja: could also not show it when its 0.

This could do with a UI rework imo, a table isn't the best way of laying this out.

Issue #237, opened 2021-01-17
Support <flakyFailure> and <flakyError> in junit xml report


ijuma:
Both Maven and Gradle support automatically re-running tests on failure and marking them as flaky if they pass on the retry. For better or worse, flaky tests are a common problem for integration and system tests. Marking such tests as flaky would be very helpful.

The [Flaky Test Handler](https://plugins.jenkins.io/flaky-test-handler/) plugin has this functionality and more, but it was last released 5 years ago and not actively maintained. It would be great if the basic reporting functionality was part of the core junit plugin.

Specifics when it comes to Maven surefire:

> flakyFailure and flakyError elements will be used in the generated xml report to include information of each failing re-runs.
https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

Similarly, Gradle:

> When mergeReruns is enabled, if a test fails but is then retried and succeeds, its failures will be recorded as <flakyFailure> instead of <failure>, within one <testcase>
https://docs.gradle.org/6.8/userguide/java_testing.html#mergereruns

timja:
Contributions welcome...

ijuma:
Looks like flaky-test-handler had a new release recently https://plugins.jenkins.io/flaky-test-handler/#releases

PR #765, opened 2025-11-17
Issue #237 parse flakyFailure and rerunFailure from surefire report files

olamy:
Part of #237

- #237

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/CaseResult.java
+++ b/src/main/java/hudson/tasks/junit/CaseResult.java
@@ -84,8 +84,8 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     private String errorStackTrace;
     private String errorDetails;
     private Map<String, String> properties;
-    private List<FlakyFailure> flakyFailures;
-    private List<RerunFailure> rerunFailures;
+    private List<Failure> flakyFailures;
+    private List<Failure> rerunFailures;
 
     @SuppressFBWarnings(value = "SE_TRANSIENT_FIELD_NOT_RESTORED", justification = "Specific method to restore it")
     private transient SuiteResult parent;
@@ -275,8 +275,8 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
         return Math.min(365.0f * 24 * 60 * 60, Math.max(0.0f, d));
     }
 
-    private static List<FlakyFailure> parseFlakyFailures(Element testCase) {
-        List<FlakyFailure> flakyFailures = new ArrayList<>();
+    private static List<Failure> parseFlakyFailures(Element testCase) {
+        List<Failure> failures = new ArrayList<>();
         List<Element> flakyFailuresElements = testCase.elements("flakyFailure");
         if (flakyFailuresElements != null) {
             for (Element flakyFailuresElement : flakyFailuresElements) {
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -84,6 +84,8 @@ public class CaseResult extends TestResult implements Comparable<CaseResult> {
     private String errorStackTrace;
     private String errorDetails;
     private Map<String, String> properties;
+    private FlakyFailure[] flakyFailures;
```
timja: any reason these are an array rather than a list?

olamy: for some reasons, using `List` is creating some empty elements in xml and fail `TestResultTest#bigResultReadWrite`

jglick: Because you are using `record`s, I think. XStream does not support them well. https://github.com/x-stream/xstream/issues/345 If you switch to `final class`es (yes it is irritating) then you should be able to make this a `List<FlakyFailure>` and XStream should produce (semi-)attractive and functional XML.

Just be sure the concrete type is `ArrayList`! Do not try to use `List.of()`, `Collections.unmodifiableList`, etc. in the field itself. You can always clone lists in a getter and/or setter or use an unmodifiable collection in a getter.

olamy: The problem was more about returning an empty liist which create an empty element in the xml and so was breaking the test `TestResultTest#bigResultReadWrite`
Now by using List but returning null instead of empty list the test is passing.

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -1040,6 +1076,14 @@ void replaceParent(SuiteResult parent) {
         this.parent = parent;
     }
 
+    public @CheckForNull List<FlakyFailure> getFlakyFailures() {
+        return flakyFailures == null ? null : Collections.unmodifiableList(flakyFailures);
+    }
+
+    public @CheckForNull List<RerunFailure> getRerunFailures() {
+        return rerunFailures == null ? null : Collections.unmodifiableList(rerunFailures);
+    }
```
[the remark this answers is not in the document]
olamy: fixed

src/main/java/hudson/tasks/junit/CaseResult.java
```diff
@@ -1040,6 +1076,14 @@ void replaceParent(SuiteResult parent) {
         this.parent = parent;
     }
 
+    public @CheckForNull List<FlakyFailure> getFlakyFailures() {
```
uhafner: The `@CheckForNull` is not required now. 

I wonder, why the field itself still is nullable? All those ugly `if (x != null)` guards could be eliminated by working with an empty collection. You can add a readResolve method to ensure that the field is set in old serializations.

src/main/java/hudson/tasks/junit/CaseResult.java
jglick: Yeah just double-check the XML format on disk by inspection.

uhafner (commented):
I'm not a maintainer of the plugin so I don't want to create too much overhead for you. 

So only a non-binding comment (as another Jenkins plugin developer):
I am wondering why the same code is used for `flakyFailures` and `rerunFailures`. Even the entities are identical. Wouldn't it simplify everything when the values are stored in an entity `Failure`? One is created for the `flakyFailures` element and one for the `rerunFailures` element?

olamy:
@uhafner yes good point. I have refactored a bit to simplify.  back to record and use a single entity.

olamy:
@timja back to review and ready

Issue #766, opened 2025-11-19
Allow sorting by package and/or name in new test failures summary page


jcelaya:
### What feature do you want to see added?

The new failures summary page shows the test name in front of the package/class name, and sorting is done by test name first. The result is that tests from different packages are grouped together:

<img width="727" height="640" alt="Image" src="https://github.com/user-attachments/assets/a77f175a-8c22-41ac-84ab-8de43ac6737d" />

It would be nice to get the previous behavior of sorting by package/class name before test name, or even allow both.

PR #1195, opened 2026-01-04
Group failed tests by package name in the test results page

Anshul-target:
## Problem
 According to [ #766](https://github.com/jenkinsci/junit-plugin/issues/766) , the test failure summary page showed all the failed cases in flat list , making it difficult for the developers to find the packages with most test failure, also tests from different packages was mixed together making it difficult to find all the failed test cases of particular package

## Solution
- Package level grouping of the test classes 
- Package are sorted lexicographically
- Package headers shows the package name and test count
- Within each package the test are sorted by the classname, then method name

## Changes

1. **TestResult.java**: Added `getFailedTestsByPackage()` method that sort the packages alphabetically
2.  **table-grouped.jelly**: New UI component that displays grouped test results with the package headers
3. **MetaTabulatedResult/body.jelly**: Modified to use the `table-grouped` component instead of flat `table` component

 

### Testing

Tested with the JUnit test reports containing multiple packages:

- Package headers display correctly with test counts
- Tests are properly grouped under their packages
- Sorting  works correctly

### Before

Screenshot 2026-01-03 130631

### After
Screenshot 2026-01-03 131715

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResult.java
+++ b/src/main/java/hudson/tasks/junit/TestResult.java
@@ -66,8 +66,9 @@ import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.StaplerRequest2;
 import org.kohsuke.stapler.StaplerResponse2;
 import org.kohsuke.stapler.export.Exported;
+import java.util.LinkedHashMap;
 
-/**
+/*
  * Root of all the test results for one build.
  *
  * @author Kohsuke Kawaguchi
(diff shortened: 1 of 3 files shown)


janfaracik:
Haven't had a chance to test this/think about the UX for this, but as a simple fix for the linked issue you could just move the `data` attribute:

image

jtnord:
> According to [ #766](https://github.com/jenkinsci/junit-plugin/issues/766) , the test failure summary page showed all the failed cases in flat list , making it difficult for the developers to find the packages with most test failure,

go to the table below the list and sort by failed, this is not difficult to do!

> also tests from different packages was mixed together making it difficult to find all the failed test cases of particular package

if you want information about a specific package then you should use the links in the table.

Now I am not saying that the "Failed tests" list is correct (I find the UX worse than it was previously) however currently I feel this PR makes it worse, and does not  improving the UX.  For example if you have 10 failures in 10 different packages this makes for a longer page with no benefit.

PR #1236, opened 2026-05-07
Add duration to TestResultSummary

### LINKED ISSUE #1235: Duration is missing from TestResultSummary
### What feature do you want to see added?

If one wants a summary of the recorded test results, including the test **duration**, one has to query the Jenkins API using for example
```
curl ${BUILD_URL}/testReport/api/json?tree=failCount,passCount,skipCount,duration
```
It would be more convenient if the `TestResultSummary` object returned by the `junit` step would also contain that duration.

### Are you interested in contributing this feature?

Yes!

HannesWell:
If one wants a summary of the recorded test results, including the tests **duration**, currently one has to query the Jenkins API via the CLI using for example
```
curl ${BUILD_URL}/testReport/api/json?tree=failCount,passCount,skipCount,duration
```
This adds that `duration` to the `TestResultSummary` object returned by the `junit` step to make it more convenient to get it.

Fixes https://github.com/jenkinsci/junit-plugin/issues/1235

### Testing done

~~Not yet, but I can extend the tests if necessary.~~
The existing test cases are extended to also cover the new `duration` field.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/hudson/tasks/junit/TestResultSummary.java
+++ b/src/main/java/hudson/tasks/junit/TestResultSummary.java
@@ -13,16 +13,22 @@ public class TestResultSummary implements Serializable {
     private int skipCount;
     private int passCount;
     private int totalCount;
+    private float duration;
 
     @Deprecated
     @Restricted(DoNotUse.class)
     public TestResultSummary() {}
 
     public TestResultSummary(int failCount, int skipCount, int passCount, int totalCount) {
+        this(failCount, skipCount, passCount, totalCount, -1.f);
+    }
+
+    public TestResultSummary(int failCount, int skipCount, int passCount, int totalCount, float duration) {
         this.failCount = failCount;
         this.skipCount = skipCount;
         this.passCount = passCount;
         this.totalCount = totalCount;
+        this.duration = duration;
     }
 
     public TestResultSummary(TestResult result) {
@@ -30,6 +36,7 @@ public class TestResultSummary implements Serializable {
         this.skipCount = result.getSkipCount();
         this.passCount = result.getPassCount();
         this.totalCount = result.getTotalCount();
+        this.duration = result.getDuration();
         if (totalCount == 0) {
             for (SuiteResult suite : result.getSuites()) {
                 if (!suite.getCases().isEmpty()) {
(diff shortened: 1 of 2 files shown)


HannesWell:
Since the solution seems relatively simple I wonder why this wasn't done yet and a separate `TestDurationResultSummary` exists? 
Is it computationally more expensive to get the duration and too expensive to do it by default?
If yes, could one solution be to explicitly request the inclusion of the duration in the returned object (via a new parameter) and return a sub-class in that case?
Or to batch the requests that compute the number of tests (passed, failed, skipped) with the duration. But I don't know the eventually used implementation of `TestResultImpl` and how they work internally.

timja:
This may be ok, see some prior objections to adding more to the summary:

https://github.com/jenkinsci/junit-plugin/pull/1196#discussion_r2683302222

https://github.com/jenkinsci/junit-plugin/pull/128#issuecomment-1188339481

cc @jglick

src/main/java/hudson/tasks/junit/TestResultSummary.java
```diff
@@ -13,23 +13,26 @@ public class TestResultSummary implements Serializable {
...
     private int passCount;
     private int totalCount;
+    private float duration;
 
     @Deprecated
     @Restricted(DoNotUse.class)
     public TestResultSummary() {}
 
-    public TestResultSummary(int failCount, int skipCount, int passCount, int totalCount) {
+    public TestResultSummary(int failCount, int skipCount, int passCount, int totalCount, float duration) {
```
timja, line 26: this would break the tests in this plugin:
https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/blob/c3f06cb528b2e9158039c3b9805065644d711c08/src/test/java/org/jenkinsci/plugins/junitrealtimetestreporter/storage/H2JunitTestResultStorage.java#L242

you generally can't safely modify a constructor, you need to add a new one and have the old one delegate to it

HannesWell: Understand, implemented it as you suggested.

HannesWell:
> This may be ok, see some prior objections to adding more to the summary:

Thanks for the links.
As far as I understood them, the concern was mainly about the memory consumption of the resulting object, not about the runtime computing it.
In that context, this change should indeed add only little overhead.

HannesWell:
@timja and @jglick, is there anything missing from my side or can we proceed here?
Thanks in advance.

timja (approved):

Issue #652, opened 2024-09-26
Memory leak related to `resultCache` in TestResultAction


l-olsem:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  ```text

Jenkins: 2.462.1
OS: Linux - 6.8.4-3-pve
Java: 21.0.2 - <company-name-reducted> S.A. (OpenJDK 64-Bit Server VM)
---
all-changes:1.5
analysis-model-api:12.4.0
antisamy-markup-formatter:162.v0e6ec0fcfcf6
apache-httpcomponents-client-4-api:4.5.14-208.v438351942757
apache-httpcomponents-client-5-api:5.3.1-110.v77252fb_d4da_5
asm-api:9.7-33.v4d23ef79fcc8
authentication-tokens:1.119.v50285141b_7e1
badge:1.13
blueocean:1.27.14
blueocean-bitbucket-pipeline:1.27.14
blueocean-commons:1.27.14
blueocean-config:1.27.14
blueocean-core-js:1.27.14
blueocean-dashboard:1.27.14
blueocean-display-url:2.4.3
blueocean-events:1.27.14
blueocean-git-pipeline:1.27.14
blueocean-github-pipeline:1.27.14
blueocean-i18n:1.27.14
blueocean-jwt:1.27.14
blueocean-personalization:1.27.14
blueocean-pipeline-api-impl:1.27.14
blueocean-pipeline-editor:1.27.14
blueocean-pipeline-scm-api:1.27.14
blueocean-rest:1.27.14
blueocean-rest-impl:1.27.14
blueocean-web:1.27.14
bootstrap5-api:5.3.3-1
bouncycastle-api:2.30.1.78.1-248.ve27176eb_46cb_
branch-api:2.1178.v969d9eb_c728e
build-blocker-plugin:166.vc82fc20b_a_ed6
buildresult-trigger:0.18
caffeine-api:3.1.8-133.v17b_1ff2e0599
checks-api:2.2.0
cloudbees-bitbucket-branch-source:888.v8e6d479a_1730
cloudbees-folder:6.942.vb_43318a_156b_2
command-launcher:115.vd8b_301cc15d0
commons-compress-api:1.26.1-2
commons-lang3-api:3.17.0-84.vb_b_938040b_078
commons-text-api:1.12.0-129.v99a_50df237f7
configuration-as-code:1850.va_a_8c31d3158b_
configuration-as-code-groovy:1.1
coverage:1.16.1
[...]

PR #1239, opened 2026-07-09
Optimize the test history some more

mdealer:
- Reinstate the link to specific test result in the test history table in place of the parent build, makes navigation in the history easier.
- Increase result cache cleanup interval, make cleanup non-blocking and avoid updating timestamp on SoftReference. Should reduce memory pressure, give GC more room to work.
- Avoid touching all the builds when about to show the history page, avoids a potentially catastrophic memory usage spike right after computing the history.
- Always show full test name in the failed tests list, make it possible to distinguish tests when viewing combined results of multiple namespaces.

Improves https://github.com/jenkinsci/junit-plugin/issues/652

### Testing done

- [x] Internal test instance 
- [x] Internal production instance - memory pressure seems significantly less from tests on my side.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/tasks/junit/History.java
+++ b/src/main/java/hudson/tasks/junit/History.java
@@ -713,6 +713,7 @@ public class History {
 
                             return new HistoryTestResultSummary(
                                     build,
+                                    resultInRun,
                                     resultInRun.getDuration(),
                                     resultInRun.getFailCount(),
                                     resultInRun.getSkipCount(),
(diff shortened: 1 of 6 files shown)


timja:
@janfaracik is your work for fixing the UX issues close? Might be better to use that instead and separate out the performance improvements?

mdealer:
> @janfaracik is your work for fixing the UX issues close? Might be better to use that instead and separate out the performance improvements?

I separated the performance improvements out into https://github.com/jenkinsci/junit-plugin/pull/1242

Nevertheless, the UX changes should be still considered. I cannot use certain areas of the history without those changes.

PR #1251, opened 2026-08-18
Replace every URL-illegal character in TestObject.safe

### LINKED ISSUE #1211: Some unsafe url characters are not replaced by TestObject.safe
### Jenkins and plugins versions report

JUnit plugin version: 1396.v095840ed8491

### What Operating System are you using (both controller, and any agents involved in the problem)?

The problem is operating system independent

### Reproduction steps

Have a look at https://github.com/jenkinsci/junit-plugin/blob/095840ed8491084d8b8af6e2ed7c744c25856ade/src/main/java/hudson/tasks/test/TestObject.java#L415C5-L432C6

There are some replacements of unsafe url characters:

Image

But in [rfc1738](https://www.[ietf.org/rfc/rfc1738.txt](https://www.ietf.org/rfc/rfc1738.txt)) there are some more unsafe url characters listed. For example:

Image

### Expected Results

All unsafe url characters should be replaced.

### Actual Results

Some unsafe url characters, for example "|", are not replaced.

### Anything else?

_No response_

### Are you interested in contributing a fix?

_No response_

piyush2676:
Fixes #1211

### Problem

`TestObject.safe()` replaced eight characters (`/ \ : ? # % < >`), but RFC 3986 forbids several more from appearing literally in a URL path segment. Everything else — `|`, a space, `[`, `]`, `{`, `}`, `^`, a backtick, `"`, control characters — was carried through verbatim into `getId()`, which `getUrl()` concatenates into the URL without encoding it:

```java
public String getUrl() {
    return getRun().getUrl() + getTestResultAction().getUrlName() + "/" + getId();
}
```

The reporter hit this with `|` in a name. The same method is used by jenkinsci/junit-attachments-plugin#201 to build a path.

### Change

`safe()` now replaces every ASCII character that RFC 3986 does not permit literally in a path segment, plus the C0 controls and DEL.

Non-ASCII characters are deliberately left alone: the browser percent-encodes them as UTF-8 so they survive the round trip, and replacing them would mangle test names that are not written in English for no benefit.

The replacement runs in a single pass and allocates nothing when the name is already safe, which also retires the chain of eight `String#replace` calls and the TODO comment that asked for exactly this.

### Compatibility

This changes the URL of any package/class/suite whose name contains one of the newly replaced characters. Those URLs were already invalid, but they are user-visible, so it is worth calling out — in particular the **space**, which is the widest-reaching one:

- `CaseResult.getSafeName()` already replaces spaces (it keeps only `Character.isJavaIdentifierPart`), so this has always been the behaviour at the case level. This change makes the package/class/suite levels agree with it.
- `TestResultTest.testStartTimes` shows the existing asymmetry: it already looks up cases as `home_button_redirects_to_home_test` while the XML says `home button redirects to home test`, but looks up classes with the spaces intact. I updated the class-level lookups in that test to match.

Happy to drop the space from the set if you would rather leave those URLs untouched.

### Testing done

- Six new unit tests in `TestObjectTest` covering every illegal character, control characters, the characters that must be preserved, non-ASCII names, null/empty, and repeated occurrences.
- `mvn test` on the full suite: 506 tests, 0 failures, 0 errors, 3 skipped.

### DIFF
--- a/src/main/java/hudson/tasks/test/TestObject.java
+++ b/src/main/java/hudson/tasks/test/TestObject.java
@@ -405,6 +405,16 @@ public abstract class TestObject extends hudson.tasks.junit.TestObject {
     private static final Map<String, Map<TestObject, Void>> UNIQUIFIED_NAMES = new HashMap<>();
 
     /**
+     * Characters that {@link #safe(String)} replaces, in addition to the space, the control
+     * characters and {@code DEL}.
+     *
+     * <p>These are the printable ASCII characters that RFC 3986 does not permit to appear
+     * literally in a URL path segment, plus {@code /} (the separator {@link #getId()} builds
+     * paths with) and {@code :} (replaced ever since this method was introduced).
+     */
+    private static final String UNSAFE_CHARS = "\"#%/:<>?[\\]^`{|}";
+
+    /**
      * Replaces URL-unsafe characters.
      * If s is an empty string, returns "(empty)" otherwise the generated URL would contain
      * two slashes one after the other and getDynamic() would fail.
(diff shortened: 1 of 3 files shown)


timja (approved):
Thanks LGTM

MarkEWaite:
Thanks for the improvement @piyush2676. The improvement is also needed in the `AttachmentPublisherTest.testAttachmentsWithStrangeFileNames` test in the junit-attachments plugin.

The plugin bill of materials weekly release detected that JUnit plugin release [1422.v580465cc5d44](https://github.com/jenkinsci/junit-plugin/releases/tag/1422.v580465cc5d44) broke that test.  Refer to my [comment](https://github.com/jenkinsci/bom/issues/7328#issuecomment-5382364146) in issue:

* https://github.com/jenkinsci/bom/issues/7328

Could you propose a pull request to the AttachmentPublisherTest in the junit attachments plugin that fixes the test?

In the meantime, I will revert JUnit plugin releases 1422.x and 1424.x from the plugin BOM because they fail the AttachmentPublisherTest in the junit attachments plugin

piyush2676:
Thanks for the detailed report and the reproduction steps, @MarkEWaite — they made this quick to track down.

Proposed fix: https://github.com/jenkinsci/junit-attachments-plugin/pull/210

It turned out not to be a test-only problem. `AttachmentPublisher.getAttachmentPath` stores attachments under `TestObject.safe(testName)`, but `TestClassAttachmentTestAction.getUrl` raw-encoded the *unsanitised* test case name, so the link only resolved for names that `safe()` happened to leave untouched:

```java
return getUrlName() + "/" + Util.rawEncode(testCase) + "/" + Util.rawEncode(filename);
```

With the wider `safe()` in 1422.v580465cc5d44, the test case `A-001 Unicode` stores its attachments in `A-001_Unicode` while the link still pointed at `A-001%20Unicode`:

```
404 Not Found for .../testReport/com.example.test/SignupTest/attachments/A-001%20Unicode/unic%C3%B6deAnd...
```

The PR extracts the sanitisation into `AttachmentPublisher.getStorageName` and builds the URL from it, so both sides derive the directory name the same way. It is backward compatible — on junit 1421 and earlier `safe()` leaves those characters alone, so the generated URL is unchanged.

Verified with your pom changes (baseline `2.541`, BOM `6904.v75208ee5a_206`, junit pinned to `1422.v580465cc5d44`): the test fails with the 404 above before the change, and `mvn -Dtest=AttachmentPublisherTest verify` is green afterwards (Tests run: 12, Failures: 0, Errors: 0). I left those pom changes out of the PR since the fix does not require a newer junit.

One related case I did not touch, in case you would rather it were handled before 1422.x goes back into the BOM: attachments captured by *older* builds are on disk under the old, less-sanitised directory names, so after an upgrade those historical builds will not find them either — `areAttachmentsStoredAtClassLevel` looks them up through the same new `safe()`. Fixing that would mean a fallback to the legacy name when the sanitised directory is missing. Happy to add it to the PR if you think it is worth carrying.

MarkEWaite:
> One related case I did not touch, in case you would rather it were handled before 1422.x goes back into the BOM: attachments captured by _older_ builds are on disk under the old, less-sanitised directory names, so after an upgrade those historical builds will not find them either — `areAttachmentsStoredAtClassLevel` looks them up through the same new `safe()`. Fixing that would mean a fallback to the legacy name when the sanitised directory is missing. Happy to add it to the PR if you think it is worth carrying.

I don't think that we need to include legacy name searching in the pull request.  Thanks for the fix!

